PR #31, opened 2023-05-31
[#4] [API] As a user, I can access some JSON data

### LINKED ISSUE #4: [API] As a user, I can access some JSON data
## Why

Set up JSON endpoint for the APIs

## Acceptance Criteria

- Create new API module
- Create endpoint `GET` `/keywords`
    - Endpoint returns JSON with `JSONAPI`
```
[
		{
			"id": "1",
			"type": "keyword",
			"attributes": {
				"name": "First keyword"
			}
		},
		{
			"id": "1",
			"type": "keyword",
			"attributes": {
				"name": "Second keyword"
			}
		}
	]
```

blyscuit:
- Close #4 

## What happened 👀

Add API endpoint returning JSONAPI. 

## Insight 📝

- Add `jsonapi-serializer`.
- Add sample controller `KeywordController`.
- Coverage now working correctly by following the one line change from this article https://medium.com/geekculture/how-to-fix-inaccurate-test-coverage-outputs-from-simplecov-f4db553e7bee.

## Proof Of Work 📹

Screen Shot 2023-06-01 at 16 11 30

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -1,0 +1,8 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ApplicationController < ActionController::API
+    end
+  end
+end
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -2,7 +2,6 @@
 
 module Api
   module V1
-    class ApplicationController < ActionController::API
-    end
+    class ApplicationController < ActionController::API; end
   end
 end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -4,7 +4,9 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        render json: KeywordSerializer.new(Keyword.new(1, 'Hello')).serializable_hash.to_json
+        keywords = [Keyword.new(1, 'First keyword'), Keyword.new(1, 'Second keyword')]
+
+        render json: KeywordSerializer.new(keywords).serializable_hash.to_json
       end
     end
   end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -1,0 +1,14 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class KeywordsController < ApplicationController
+      def index
+        keyword = Keyword.new
+        keyword.id = 1
+        keyword.name = "Hello"
+        render json: KeywordSerializer.new(keyword).serializable_hash.to_json
+      end
+    end
+  end
+end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -4,12 +4,8 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        first_keyword = Keyword.new
-        first_keyword.id = 1
-        first_keyword.name = 'First keyword'
-        second_keyword = Keyword.new
-        second_keyword.id = 2
-        second_keyword.name = 'Second keyword'
+        first_keyword = Keyword.new({ id: 1, name: 'First keyword' })
+        second_keyword = Keyword.new({ id: 2, name: 'Second keyword' })
 
         render json: KeywordSerializer.new([first_keyword, second_keyword]).serializable_hash.to_json
       end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -4,10 +4,7 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        keyword = Keyword.new
-        keyword.id = 1
-        keyword.name = 'Hello'
-        render json: KeywordSerializer.new(keyword).serializable_hash.to_json
+        render json: KeywordSerializer.new(Keyword.new(1, 'Hello')).serializable_hash.to_json
       end
     end
   end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -6,7 +6,7 @@ module Api
       def index
         keyword = Keyword.new
         keyword.id = 1
-        keyword.name = "Hello"
+        keyword.name = 'Hello'
         render json: KeywordSerializer.new(keyword).serializable_hash.to_json
       end
     end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -4,9 +4,14 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        keywords = [Keyword.new(1, 'First keyword'), Keyword.new(1, 'Second keyword')]
+        first_keyword = Keyword.new
+        first_keyword.id = 1
+        first_keyword.name = 'First keyword'
+        second_keyword = Keyword.new
+        second_keyword.id = 2
+        second_keyword.name = 'Second keyword'
 
-        render json: KeywordSerializer.new(keywords).serializable_hash.to_json
+        render json: KeywordSerializer.new([first_keyword, second_keyword]).serializable_hash.to_json
       end
     end
   end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,5 +1,10 @@
 # frozen_string_literal: true
 
 class Keyword
-  attr_accessor :id, :name
+  attr_reader :id, :name
+
+  def initialize(id, name)
+    @id = id
+    @name = name
+  end
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
-class Keyword 
+class Keyword
   attr_accessor :id, :name
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,5 +1,7 @@
 # frozen_string_literal: true
 
 class Keyword
+  include ActiveModel::API
+
   attr_accessor :id, :name
 end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,0 +1,5 @@
+# frozen_string_literal: true
+
+class Keyword 
+  attr_accessor :id, :name
+end
--- a/app/models/keyword.rb
+++ b/app/models/keyword.rb
@@ -1,10 +1,5 @@
 # frozen_string_literal: true
 
 class Keyword
-  attr_reader :id, :name
-
-  def initialize(id, name)
-    @id = id
-    @name = name
-  end
+  attr_accessor :id, :name
 end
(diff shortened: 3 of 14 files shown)


app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,14 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class KeywordsController < ApplicationController
+      def index
+        keyword = Keyword.new
+        keyword.id = 1
+        keyword.name = 'Hello'
```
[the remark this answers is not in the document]
blyscuit: This will be removed later so can ignore.

spec/models/keyword_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Keyword, type: :model do
+  describe 'Init' do
+    let(:keyword) { Fabricate(:keyword) }
```
byhbt:
```suggestion
```

Same with the comment here: https://github.com/blyscuit/rails-ic/pull/32/files#r1217354441

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,20 @@
...
+
+RSpec.describe Api::V1::KeywordsController, type: :controller do
+  describe 'GET#index' do
+    it 'returns expected status' do
+      get :index
+
+      expect(response).to have_http_status(:success)
+    end
+
+    it 'returns response with valid keyword' do
```
byhbt: We can use different verb to make the sentence more concise.

```suggestion
    it 'responds the valid keyword' do
```

app/controllers/api/v1/application_controller.rb
```diff
@@ -0,0 +1,8 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class ApplicationController < ActionController::API
+    end
```
malparty:
```suggestion
    class ApplicationController < ActionController::API; end
```

Since ruby 3 we can inline empty classes :) it doesn't change much, but just to let you know about it

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class KeywordsController < ApplicationController
+      def index
+        render json: KeywordSerializer.new(Keyword.new(1, 'Hello')).serializable_hash.to_json
```
malparty: As this is an `index` action, should we implement the right format directly? 
If yes, we can just serialize a list of keywords:

```suggestion
        keywords = [Keyword.new(1, 'First keyword'), Keyword.new(1, 'Second keyword')]
        
        render json: KeywordSerializer.new(keywords).serializable_hash.to_json
```

malparty: Another option is to use the `show` action for this story, and to keep single item serialization ;-)

app/models/keyword.rb
```diff
@@ -0,0 +1,10 @@
+# frozen_string_literal: true
+
+class Keyword
+  attr_reader :id, :name
+
+  def initialize(id, name)
+    @id = id
+    @name = name
+  end
```
malparty:
```suggestion
  attr_accessor :id, :name
```

Does it work with just `attr_accessor`? (i.e. getter and setter)

config/environments/test.rb
```diff
@@ -14,7 +14,7 @@
   # Eager loading loads your whole application. When running a single test locally,
   # this probably isn't necessary. It's a good idea to do in a continuous integration
   # system, or in some way before deploying your code.
-  config.eager_load = ENV["CI"].present?
+  config.eager_load = false
```
malparty: Can you provide more insight why this change?

In general, we prefer to keep `ENV["CI"].present?` because:
- In local test run, we want tests to run as fast as possible, so not loading the whole code of the application allow to only load what the tests required and run faster
- In CI, we generally run the whole test suite, so there is no time gained by using lazy loading. However, using `eager loading` will ensure to raise an error/exception in case a part of the code is corrupted (even if there is no test running on that part of the code). So it allow to identify an error BEFORE the error is raised in the staging/production environment 🙌 

For that reason, if you had an issue with this, we might want to fix the issue and keep eager loading for CI 🤔 (did you add a `CI` env variable in your GitHub action env variables?)

spec/fabricators/keyword_fabricator.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+Fabricator(:keyword) do
+  on_init { init_with(FFaker::Identification.ssn, FFaker::FreedomIpsum.word) }
```
malparty:
```suggestion
  id { init_with(FFaker::Identification.ssn }
  name { FFaker::FreedomIpsum.word) }
```

If you use the attr_accessor as suggested above, then you can use a simpler and more explicit Fabricator ;-)

spec/models/keyword_spec.rb
```diff
@@ -0,0 +1,13 @@
...
+
+    it 'has name' do
+      expect(keyword.name).not_to be_nil
+    end
```
malparty: I would not recommend testing the attributes of an object.
The main reason is that attributes are a mean to achieve something, but they are not an end. 
It is a bit similar for the "model associations", we could write beautiful tests to assert the one-to-many or many-to-many associations between models, but the tests would be similar to the code in the model (see [Tautological Tests](https://randycoulman.com/blog/2016/12/20/tautological-tests/)), so it is better to have tests that focus on the logic and that relies on these associations to test some business logic.


In case you really want to test it, I would recommend going 1 step further to be closer to testing business logic:

```suggestion
    it 'exposes its name' do
      keyword = Fabricate(:keyword, name: 'A keyword name')
      
      expect(keyword.name).to eq('A keyword name')
    end
```

but again, my opinion is that it is not needed 😇

malparty (changes requested):

app/models/keyword.rb
blyscuit: It works with `attr_accessor` but lint catches this because we are not setting the values, so `attr_reader` to limit the write access to these values sound better for me.

spec/fabricators/keyword_fabricator.rb
blyscuit: I prefer to keep using `attr_reader` for this to satisfy the line, I can come back to look at this if this model requires change to `attr_accessor` 🙏

spec/models/keyword_spec.rb
blyscuit: I agree with you, this test is redundant of the init functionality of the object but I would like to keep it here so that the `keyword` model has some test that I can build on in the future features. I think of it as a starting point for other tests for this model.

Taking you suggestion, I will update the test to not use Fabricator for the `expose` test, as we are testing the `init` function and also the `read` permission.

spec/models/keyword_spec.rb
blyscuit: Thank you very much, it seems I updated this in the wrong PR.

app/controllers/api/v1/keywords_controller.rb
blyscuit: Sounds good, we can have it in the future-correct format. I will update the story.

config/environments/test.rb
blyscuit: Thank you for the detail explanation. I think I found some weird issues when using eager load on GH Actions but now it is resolved. Reverting this code back. 🙏

app/models/keyword.rb
malparty: Lint is here to help us pay attention to the right things. But it is not here to lower the quality of the code or to make Fabricator more complex than what they should be 😇 

`keywords` aims to be an ActiveRecord model in the future. To have the least need for changes, `attr_accessor` will be used by the active record and therefore the way all other libraries are optimized (e.g. Fabrication).

While it's `working` with attr_reader and an initializer, it makes tests less standard and it will require you to refactor the code more later. That's the reason why I recommend using `attr_accessor` in this case.

spec/models/keyword_spec.rb
malparty: Ok. 

Small note: `read permission` => It's usually called `visibility` (attributes visibility, method visibility, ... Either public, protected or private).

Let's keep `permission` for requests params and authorization ;-)

db/schema.rb
malparty, line 1: Before merging this PR, please remove this change 🙏 

The schema file is autogenerated, but its changes must always reflect the DB migrations added in the same PR. These changes come from another PR. 😇

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::KeywordsController, type: :controller do
+  describe 'GET#index' do
+    it 'returns expected status' do
```
```suggestion
    it 'returns the success status' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -0,0 +1,21 @@
...
+      get :index
+
+      expect(response).to have_http_status(:success)
+    end
+
+    it 'responds the valid keywords' do
```
```suggestion
    it 'returns the valid keywords' do
```

(for consistency with the previous test — both information come from the `response` object ;) )

malparty (approved):
Approved ahead with some small need for changes (mainly the schema file to rollback).

While I recommend using attr_accessor, it's not a blocker either and future refactoring might help better understand the way other libraries optimize for the `ActiveModel` code structure. So all good with this! :)

db/schema.rb
blyscuit: Good catch.

app/models/keyword.rb
blyscuit: I see, thank you for the insight, I will update this model to `attr_accessor` to be future proof.

byhbt (approved):

app/models/keyword.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+class Keyword
+  attr_accessor :id, :name
```
malparty, line 6:
```suggestion
  include ActiveModel::Model

  attr_accessor :id, :name
```

Magic happens when you include `include ActiveModel::Model` 🚀 

You can then use it this way in the controller:

```
      def index
        first_keyword = Keyword.new({ id: 1, name: 'Test' })
        second_keyword = Keyword.new({ id: 2, name: 'Test' })
```

malparty: Recommended reading about this: https://guides.rubyonrails.org/active_model_basics.html

blyscuit: How do we decide between `ActiveModel::API` and `ActiveModel::Model`? From my understanding, Rail 7 forward, `ActiveModel::API` is preferred?

malparty: It seems like currently they have the exact same behavior as ActiveModel includes ActiveAPI. But in the future, ActiveModel should have more "ActiveRecord like" features, while ActiveAPI aims to stay minimalist and focused on the View helpers (i.e. not including DB related helpers).

In this case, both are fine as in the future PR we will replace this `include ActiveModel/API` with a full ActiveReccord inheritance. ;-)

malparty (approved):

PR #32, opened 2023-06-01
[#5] [Backend] Set up Authentication

### LINKED ISSUE #5: [Backend] Set up Authentication
## Why

Set up authentication with `Devise` to accommodate `oauth` feature.

## Acceptance Criteria

- Use `Devise`, create endpoint with authentication locked and non-locked.
    - `GET` `/private_items` is authentication locked
    - `POST` `/public_items` is non-locked.
- Create a Chore task to remove the test endpoints.

blyscuit:
- Close #5 

## What happened 👀

Add Authentication to the app. Provides example for authenticated endpoints and non-authenticated endpoints.

## Insight 📝

- Set up `Devise`.
- Includes two controllers for testing purposes:
  - `PrivateItemsController`
  - `PublicItemsController`
- Create story to remove temporary endpoints https://github.com/blyscuit/rails-ic/issues/34.

## Proof Of Work 📹

Screen Shot 2023-06-01 at 18 10 35


spec/requests/private_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+        get :index
+
+        expect(response).to have_http_status(:found)
+      end
+    end
+
+    context 'when user is sign in' do
+      before(:each) do
+        sign_in Fabricate(:user)
+      end
```
byhbt: Please check this convention: https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#fixtures-isolation

We don't use the `before` and `after` when writing a spec.

spec/requests/public_controller_spec.rb
```diff
@@ -0,0 +1,13 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PublicController, type: :controller do
+  describe 'GET#index' do
+    it 'returns expected status' do
```
byhbt:
```suggestion
    it 'returns success status' do
```

spec/models/user_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe User, type: :model do
+  let(:user) { Fabricate(:user) }
```
byhbt, line 6: Same as the comment below https://github.com/blyscuit/rails-ic/pull/32/files#r1217352048

Please check the `spec/models/keyword_spec.rb` as well.

config/initializers/devise.rb
```diff
@@ -0,0 +1,313 @@
...
+
+  # ==> Controller configuration
+  # Configure the parent class to the devise controllers.
+  # config.parent_controller = 'DeviseController'
+
+  # ==> Mailer Configuration
+  # Configure the e-mail address which will be shown in Devise::Mailer,
+  # note that it will be overwritten if you use your own mailer class
+  # with default "from" parameter.
+  config.mailer_sender = 'please-change-me-at-config-initializers-devise@example.com'
```
malparty:
```suggestion
  config.mailer_sender = ENV.fetch('MAILER_SENDER')
```

app/controllers/private_controller.rb
malparty, line 1: Why not using the `keywords` controller directly? 💡

spec/fabricators/user_fabricator.rb
```diff
@@ -0,0 +1,7 @@
+# frozen_string_literal: true
+
+Fabricator(:user) do
+  email { FFaker::Internet.email }
```
malparty:
```suggestion
  email { FFaker::Internet.unique.email }
```

This will avoid getting duplicate email if in a test you need to Fabricate multiple users 👍

spec/requests/private_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PrivateController, type: :controller do
```
malparty:
```suggestion
RSpec.describe PrivateController, type: :request do
```

spec/requests/private_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+  describe 'GET#index' do
+    context 'when user is not sign in' do
+      it 'returns redirect status' do
+        get :index
+
+        expect(response).to have_http_status(:found)
+      end
```
malparty:
```suggestion
      it 'redirects to the new user session page' do
        get :index

        expect(response).to have_http_status(:found)
        expect(response).to redirect_to(new_user_session_path)
      end
```

(the `new_user_session_path` might be different in your project, check at `rails routes` CLI to see what path is expected) :)

spec/requests/private_controller_spec.rb
```diff
@@ -0,0 +1,27 @@
...
+    end
+
+    context 'when user is sign in' do
+      before(:each) do
+        sign_in Fabricate(:user)
+      end
+
+      it 'returns expected status' do
```
malparty:
```suggestion
      it 'returns success status' do
```

Let's try to avoid `expected`, `correct`, `calculated` ... words in a test description. Tests should "document" the code and so when reading the test description, a developer should understand exactly what the code is expect to do.

E.x. "given 2 coffee at 1.5USD each", "it returns the price of 3USD"
=> By just reading these 2 simple descriptions, you have a clear understanding that the method calculate the price addition of a list of coffee based on their price. ;-)
If I just say "it returns the correct price", the calculation becomes "implicit" and you have to read the test implementation to understand the code.

spec/requests/private_controller_spec.rb
blyscuit: Thank you for pointing this out. Could I ask for more clarification on the why of this conventions, is it because most test files are big and having the `before` `after` make the tests more unclear?

app/controllers/private_controller.rb
blyscuit: At this point I still want `keywords` controller to work without authentication(auth) because the goal of this project is to use `oauth` for the authentication hence devise setup here is not the completion of the auth feature.
Also having one endpoints working without auth makes testing early part of the project easier for dev. 🙏

spec/requests/private_controller_spec.rb
blyscuit: Good insight from this. I modified all `controller` specs to use `request` which is more favored than the old `controller`. In the process I had to add `config.hosts` as described by this [thread](https://stackoverflow.com/a/60573259/3736373). 🙏 

Please help me check if I understand `request` tests correctly.

spec/requests/private_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe PrivateController, type: :request do
+  describe 'GET#index' do
+    context 'when user is not sign in' do
```
malparty:
```suggestion
    context 'when the user is not signed in' do
```

spec/requests/private_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+      it 'redirects to the new user session page' do
+        get '/private'
+
+        expect(response).to have_http_status(:found)
+        expect(response).to redirect_to(new_user_session_path)
+      end
+    end
+
+    context 'when user is sign in' do
```
malparty:
```suggestion
    context 'when the user is signed in' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -2,16 +2,17 @@
 
 require 'rails_helper'
 
-RSpec.describe Api::V1::KeywordsController, type: :controller do
+RSpec.describe Api::V1::KeywordsController, type: :request do
   describe 'GET#index' do
-    it 'returns expected status' do
-      get :index
+    it 'returns success status' do
+      get '/api/v1/keywords'
```
```suggestion
      get keywords_path
```

You can see all the path and url helpers in the documentation: https://guides.rubyonrails.org/routing.html#path-and-url-helpers

(by the way, the whole page is quite interesting — I'm still learning new things each time I re-read it 🙈 )

spec/requests/private_controller_spec.rb
byhbt: yeah, tbh i am not propose this convention, and not sure the correct main idea behind it. 😅  But i think to avoid the global definition in when writing a spec. (as the heading is `Fixtures isolation`)

Every `single context` should define the related fixtures (test data) within itself. We don't have to look up up and down when debugging a specific test, no surprises 😮 It's not actually DRY however i think this is a good trade-off.

cc @malparty what is you idea on this?

malparty (approved):
Approved with some last suggestions~

byhbt (approved):

spec/requests/api/v1/keywords_controller_spec.rb
blyscuit: I am not able to get this working 🤔 Perhaps the project is missing some plugin? 
Screen Shot 2023-06-06 at 17 10 35

blyscuit: Working now with `api_v1_keywords_path` thank you for the suggestions and the link.

blyscuit: Do you think it should work with `private_path` and `public_path` as well? I tried but failed 🤔

malparty: `private` and `public` paths do not follow Rails standards — that also a reason why I recommend using real controllers over "temporary private/public" ones. 

The route is declared as a `resources route` but `private` and `public` are singular words. If you want them to work, you need to follow the Rails conventions:

- Pluralized controller name: `privates_controller.rb`
- Pluralized resources route: `resources :privates, only: :index`

Only from that Rails will be able to reflect the expected names (as far as I remember a model class is not mandatory — but to be confirmed). 


On the "idiomatic" side, `private` and `public` are not resource names (you will never have a DB table called `privates` or `publics`). If you want to follow the conventions, it might be better to choose a name like "private_item" (with "private_items_controller.rb" / and the route `resources :private_items`).

malparty: this is the "weird", "nice" and sometimes "annoying" part of Rails :D 

This famous "convention over configuration": If you follow the conventions, a lot of magic will happen (e.g. `_path` short hands, routes, model reflextion, etc...) but it means we have to be very careful in learning and applying those conventions (via Rails guides mainly) ;-)

blyscuit: Thank you very much. I am getting started to understand Ruby's magic. Updating this PR to make it more confirming to the conventions 🙏

malparty (approved):

spec/requests/private_items_controller_spec.rb
```diff
@@ -0,0 +1,26 @@
...
+require 'rails_helper'
+
+RSpec.describe PrivateItemsController, type: :request do
+  describe 'GET#index' do
+    context 'when the user is not signed in' do
+      it 'redirects to the new user session page' do
+        get private_items_path
+
+        expect(response).to have_http_status(:found)
+        expect(response).to redirect_to(new_user_session_path)
```
byhbt, line 12: For the API, i think we don't need to assert the `redirect_to` path. Because we just need the status for the client process the next step right?

```suggestion
```

blyscuit: I agree this is unnecessary, it will be removed when clean up none-API paths 🙏

PR #33, opened 2023-06-01
[#6] [Backend] Set up Authenticated API 

### LINKED ISSUE #6: [Backend] Set up Authenticated API
## Why

Set up authenticated API endpoints to allow only some endpoints to be accessiable without a Token.

## Acceptance Criteria

- Set up `Doorkeeper`.
- Create endpoints with authentication locked.
    - `GET` `/api/v1/private/sample` is authentication locked
- Create a Chore task to remove the test endpoints.

## Resources

https://rubyyagi.com/rails-api-authentication-devise-doorkeeper/

blyscuit:
- Close #6

## What happened 👀

Set up endpoints for oauth token.

## Insight 📝

- Set up `doorkeeper`
- Add `Api::V1:PrivateController` for testing purpose.

## Proof Of Work 📹

Screen Shot 2023-06-01 at 18 10 35


spec/requests/api/v1/private_controller_spec.rb
```diff
@@ -0,0 +1,15 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::PrivateController, type: :controller do
+  describe 'GET#index' do
+    context 'when user is not sign in' do
+      it 'returns unauthorized error' do
```
byhbt, line 8:
```suggestion
      it 'returns unauthorized status' do
```

byhbt:
```suggestion
    context 'when the user is unauthenticated' do
      it 'responds unauthorized status' do
```

Or `given the user does NOT logged in`, or you can propose your own.

app/controllers/api/v1/private_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class PrivateController < ApplicationController
+      def sample
```
byhbt: The method name in the controller should follow the Rails Controller method names. We just use the different name in some special case (`download`, ...). I got your point here this is just a sample, however let use the Rails way naming https://guides.rubyonrails.org/action_controller_overview.html#what-does-a-controller-do-questionmark

```suggestion
      def index
```

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -3,6 +3,7 @@
 module Api
   module V1
     class KeywordsController < ApplicationController
+      skip_before_action :doorkeeper_authorize!
       def index
```
byhbt, line 8:
```suggestion

      def index
```

spec/requests/api/v1/private_controller_spec.rb
blyscuit: I will change to `returns unauthorized status`, keeping `returns` to make it consistent with other specs 🙏

config/routes.rb
```diff
@@ -1,9 +1,14 @@
...
+  use_doorkeeper do
+    skip_controllers :authorizations, :authorized_applications, :tokens, :token_info
+  end
+
   devise_for :users
   
   namespace :api do
     namespace :v1 do
       resources :keywords, only: :index
+      get 'private/sample', to: 'private_items#index'
```
malparty: Is there a reason why we use a custom route? 👀 
While it is not forbidden, we generally try to stay within the basic resources routes, unless there is a specific need that cannot use them (e.g. a download file action that is not specific to a single `id` – so cannot use the `show` default route).

Note that this is not a big issue, but more a small practice to stay as close as possible to the Rails way ;-)

config/initializers/doorkeeper.rb
```diff
@@ -0,0 +1,18 @@
+# frozen_string_literal: true
+
+Doorkeeper.configure do
+  orm :active_record
+
```
malparty, line 5:
```suggestion
  api_only

```

According to the documentation there is an [api_only](https://doorkeeper.gitbook.io/guides/ruby-on-rails/api-mode) mode. 🤔 Did you try it? :)

> Note that I never worked with DoorKeeper, so this suggestion might not work 😇

malparty (changes requested):

app/controllers/api/v1/private_items_controller.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class PrivateItemsController < ApplicationController
```
byhbt, line 5: Instead of creating a `PrivateItemsController` for testing/mockup, i think we also could define a private controller that potentially will be used in the next tickets, so we don't need to removed it after this PR.

byhbt (approved):

app/controllers/api/v1/private_items_controller.rb
blyscuit: I think the method will be temporary, even with a feature-ready controller, I find it easier to keep the unused soon-to-be-removed code in the same place for easier clean up 🙏

config/initializers/doorkeeper.rb
blyscuit: I tried adding this flag but `rails routes` still returns `applications` routes so I still have to skip them. 🙏

malparty: While routes might still be present, I understand from the documentation that using the `api_only` flag can reduce the features loaded into the project (like views, view helpers, etc...). So even if it feels like it does not change anything, it might still be useful for the project :)

malparty (approved):
# 🚀

app/controllers/api/v1/private_items_controller.rb
byhbt: that's fine.

byhbt (approved):

PR #37, opened 2023-06-07
[#7] [API] As a user, I can Sign In and receive a JWT

### LINKED ISSUE #7: [API] As a user, I can Sign In and receive a JWT
## Why

User should be able to sign in and receive a JWT token with refresh token for use with other API.

## Acceptance Criteria

- Create endpoint `POST` `/api/v1/signin` for sign in.
    - Input parameters:
    
    ```jsx
    {
    	"grant_type": "password",
    	"email": "me@nimblehq.co",
    	"password": "123456",
    	"client_id": "{{ _.client_id }}",
    	"client_secret": "{{ _.client_secret }}"
    }
    ```
    
    - Output:
    
    ```jsx
    {
    	"access_token": "",
    	"token_type": "Bearer",
    	"expires_in": 7200,
    	"refresh_token": "",
    	"created_at": 1684995376
    }
    ```
    

## Resources

https://github.com/doorkeeper-gem/doorkeeper/blob/main/app/controllers/doorkeeper/tokens_controller.rb

blyscuit:
- close #7

## What happened 👀

Add `login` route. User can sign in with path `/api/v1/login` and received jwt token.

## Insight 📝

- Add empty `TokensController` to support custom route inside `api/v1` module.
- Add custom route for signin.
- Add `OAuthHelpers` for helping with auth related specs.

## Proof Of Work 📹
Screen Shot 2023-06-08 at 14 49 49


config/initializers/doorkeeper.rb
```diff
@@ -2,6 +2,8 @@
 
 Doorkeeper.configure do
   orm :active_record
+  api_only
```
phongvhd93: This duplicate with line number 8

spec/routing/tokens_controller_spec.rb
```diff
@@ -0,0 +1,11 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::TokensController, type: :routing do
```
byhbt: As far as i know, i most our Ruby on Rails project, we don't write test for the route.

spec/fabricators/application_fabricator.rb
```diff
@@ -0,0 +1,5 @@
+# frozen_string_literal: true
+
+Fabricator(:application, from: Doorkeeper::Application) do
+  name 'iOS client'
```
byhbt: We also can provide more variant of data by this way.

```suggestion
  name ['iOS client', 'Android client'].sample
```

spec/requests/api/v1/tokens_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+
+require 'rails_helper'
+
+RSpec.describe Api::V1::TokensController, type: :request do
+  describe 'POST#login' do
+    context 'given valid credentials' do
+      it 'returns an access token' do
+        post api_v1_login_path, params: token_request_params
+
+        expect(JSON.parse(response.body)['token_type']).to eq('Bearer')
```
byhbt, line 11: We also can use this gem `rspec-json_expectations` to assert the JSON content. Reference: https://github.com/waterlink/rspec-json_expectations

Current approach is also good enough for this case.

config/routes.rb
```diff
@@ -7,8 +7,13 @@
   
   namespace :api do
     namespace :v1 do
+      use_doorkeeper do
+        controllers tokens: 'tokens'
+        skip_controllers :authorizations, :applications, :authorized_applications, :token_info, :tokens
+      end
       resources :keywords, only: :index
       resources :private_items, only: :index
+      post '/login', to: 'tokens#create'
```
malparty: As usual, unless there is a specific need from the requirements or any technical constraints, we recomend sticking with the Rails standard.
This applies with routes: `tokens` is a resource, so we can use the `resources` short hand:
```suggestion
      resources :tokens, only: [:create]
```

Then we will have the route `POST /tokens` created (instead of `POST /login`).

spec/requests/api/v1/tokens_controller_spec.rb
blyscuit: The documentation for this library is a broken link 😢 I also tried a few of alternatives and found that the current approach is already readable.

We can revisit this library again later if the document is back.

config/routes.rb
blyscuit: @phongvhd93 You might want to do the same on your PR 🙏

byhbt (approved):

spec/requests/api/v1/tokens_controller_spec.rb
```diff
@@ -0,0 +1,48 @@
...
+      expect(JSON.parse(response.body)['token_type']).to eq('Bearer')
+    end
+  end
+
+  context 'when the user refreshes an incorrect refresh token' do
+    it 'returns an invalid_grant error' do
+      post api_v1_tokens_path, params: token_refresh_params('wrong_token')
+
+      expect(JSON.parse(response.body)['error']).to eq('invalid_grant')
+    end
```
malparty, line 65: Should we also test the status code in both the happy path and this error path? 🤔

malparty (approved):
Approved with a last recommendation :)

phongvhd93 (approved):

spec/requests/api/v1/tokens_controller_spec.rb
blyscuit: Updated with status code tests.

PR #38, opened 2023-06-09
[#8] [API] As a user, I can Register and receive a JWT

### LINKED ISSUE #8: [API] As a user, I can Register and receive a JWT
## Why

User should be able to register and receive a JWT token with refresh token for use with other API. Use can go through register flow without the need to login again.

## Acceptance Criteria

- Create endpoint `POST` `/api/v1/register`.
    - Input parameters:
    
    ```jsx
    {
    	"email": "me2@nimblehq.co",
    	"password": "123456",
         "password_confirmation": "123456"
    	"client_id": "{{ _.client_id }}"
    }
    ```
    
    - Output:
      - Http Code 201 (created)
      - Send a verification email

    

## Resources

https://rubyyagi.com/rails-api-authentication-devise-doorkeeper/

phongvhd93:
- Close #8 

## What happened 👀

Add a new route `api/v1/signup` for users to sign up with email and password.

## Insight 📝

- Add `RegistrationsController` and `RegistrationForm` to handle sign-up routes and sign-up logic (validation and save the new user to the database)
- Write tests for `RegistrationController` and `RegistrationForm` 

## Proof Of Work 📹

Screenshot 2023-06-12 at 14 02 47


spec/requests/api/v1/registrations_controller_spec.rb
```diff
@@ -0,0 +1,77 @@
...
+        post api_v1_signup_path, params: confirm_password_not_matched_signup_params
+
+        expect(response).to have_http_status(:unprocessable_entity)
+      end
+    end
+  end
+
+  private
+
+  def valid_signup_params
```
byhbt: Usually we define these params as a variable within the test instead of using a method to return like this. 💭

config/locales/en.yml
```diff
@@ -31,3 +31,7 @@
 
 en:
   hello: "Hello world"
+  error:
+    invalid_email: "Invalid email address"
+    invalid_password: "Invalid password"
+    invalid_client: "Invalid client"
```
byhbt:
```suggestion
    invalid_email: Invalid email address
    invalid_password: Invalid password
    invalid_client: Invalid client
```

app/serializers/authentication_serializer.rb
```diff
@@ -0,0 +1,3 @@
+class AuthenticationSerializer: ApplicationSerializer
+
+end
```
byhbt: We can define it within 1 line 

```suggestion
class AuthenticationSerializer: ApplicationSerializer; end
```

app/forms/registration_form.rb
```diff
@@ -0,0 +1,29 @@
...
+    super
+    @client_app = Doorkeeper::Application.find_by(uid: sign_up_params[:client_id])
+    allowed_params = sign_up_params.except(:client_id)
+    @user = User.new(allowed_params)
+  end
+
+  def save
+    unless validate_client_id && @user.valid?
+      errors.add(:user, @user.errors.full_messages.last) unless @user.errors.empty?
+      return
```
byhbt: Always add one empty line before the early `return` like this
```suggestion

      return
```

app/forms/registration_form.rb
```diff
@@ -0,0 +1,29 @@
...
+    end
+
+    @user.save
+  end
+
+  private
+
```
byhbt, line 19: We can define the `client_app` as an attr_reader like this, it's same purpose as instance variable `@` however it's clearer when we read the class. 

```
attr_reader client_app
```

In the initializer, you can init it like current approach

```
@client_app = Doorkeeper::Application.find_by(uid: sign_up_params[:client_id])
```

However in the method below, `validate_client_id`, we can ignore the `@`.

```suggestion
def validate_client_id
    return true if client_app

    errors.add(:client_id, I18n.t('doorkeeper.errors.messages.invalid_client'))
  end
```

Or even better

```ruby
  def validate_client_id
    errors.add(:client_id, I18n.t('doorkeeper.errors.messages.invalid_client')) unless client_app
  end
```

app/forms/registration_form.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class RegistrationForm < ApplicationForm
+  attr_accessor :email, :password, :password_confirmation, :client_id
```
byhbt:
```suggestion
```

We don't need this definition right? 🤔

app/controllers/api/v1/registrations_controller.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class RegistrationsController < ApplicationController
+      skip_before_action :doorkeeper_authorize!
+
+      def create
+        validation = RegistrationForm.new(sign_up_params)
```
byhbt: Follow the convention, we will pass the param to the `save(params)` instead of the `initializer` method

byhbt (changes requested):

.gitignore
```diff
@@ -42,3 +42,4 @@
 
 # Ignore Spec Cache
 spec/examples.txt
+vendor
```
malparty, line 45: Can you tell us more about where this `vendor` folder comes from? 👀

Gemfile
```diff
@@ -37,6 +37,7 @@ group :development do
   gem 'spring' # Spring speeds up development by keeping your application running in the background.
   gem 'spring-commands-rspec' # This gem implements the rspec command for Spring.
   gem 'spring-watcher-listen', '2.0.1' # Makes Spring watch the filesystem for changes using Listen
+  gem "debug"
```
malparty:
```suggestion
  gem 'debug'
```

There are a few discussions if we should not just use `"` everytime instead of `'` when string without string interpolation 🤔 
But as for now, the convention remains using single quote whenever possible 😇

malparty: Also, if you run the project with `rails s -p 3000 -b '0.0.0.0'` (when not in need to watch css or migrate the db), you can use `binding.pry` anywhere in the code (even in tests) and then have a quite nice CLI interface for step by step debug & calling any method/variable in the same way you would do in the `rails console` ;-)

app/controllers/api/v1/registrations_controller.rb
malparty:
```suggestion
        registration_form = RegistrationForm.new
```

Also let's name it `regiastration_form` or `form` ;-)

app/controllers/api/v1/registrations_controller.rb
```diff
@@ -0,0 +1,24 @@
...
+        if validation.save
+          render success: true, status: :created
+        else
+          render json: { error: validation.errors.full_messages.to_sentence }, status: :unprocessable_entity
+        end
+      end
+
+      private
+
+      def sign_up_params
```
malparty:
```suggestion
      def create_params
```

params are usually related to a specific action, so reusing the name of this action makes it easier to understand the context ;-)

app/controllers/concerns/error_handler.rb
```diff
@@ -0,0 +1 @@
+# frozen_string_literal: true
```
malparty: This file seems empty, should we remove it? :)

app/forms/registration_form.rb
malparty: Currently it was needed for the `initialize` method that acts as a Model.

But as Form objects must be initialized with the models (or no attributes) and must call `save` with the params, they won't be needed.

app/models/user.rb
```diff
@@ -6,6 +6,10 @@ class User < ApplicationRecord
   devise :database_authenticatable, :registerable,
          :recoverable, :rememberable, :validatable
 
+  validates :password, confirmation: true
+  validates :password_confirmation, presence: true
+  validates :email, format: URI::MailTo::EMAIL_REGEXP
```
malparty: As Devise already includes an email regex, it is recommended to reuse the same (to avoid some edge cases when 1 regex allow `x@y.z` when the other regex would require at least 2 chars before the `@` :harold:

You can get the Devise regex with `Devise.email_regexp`.
This regex is configured in `engines/ewa_user/config/initializers/devise.rb` with `config.email_regexp` ;-)

app/serializers/application_serializer.rb
```diff
@@ -0,0 +1,4 @@
+class ApplicationSerializer
+  include JSONAPI::Serializer
+  include Lookup
```
malparty: Do we need `Lookup` in every single serializers? 
If not, we might prefer to only include it in the serializer that will need it.

config/routes.rb
```diff
@@ -4,16 +4,17 @@
...
     namespace :v1 do
       use_doorkeeper do
         controllers tokens: 'tokens'
         skip_controllers :authorizations, :applications, :authorized_applications, :token_info, :tokens
       end
+
       resources :keywords, only: :index
       resources :private_items, only: :index
-      resources :tokens, only: [:create]
+      post '/signup', to: 'registrations#create'
```
malparty:
```suggestion
      resources :registrations, only [:create]
```

As said earlier, unless specifications or client constrains the project to use a custom route, when working with Rails we have a lot to gain if we stick to the default resources routes. Because Rails mindset is `convention over configuration`, if you follow the conventions (routes must follow the model name, controller/actions, only `index/new/edit/create/update/destroy` actions, etc...), then most of the configurations will be done implicitly and all the helpers/tools will work seamlessly. 

But at the moment you start getting "off the rails", you might have a cascade of unexpected complexity that arise :D 

Of course, when needed everything is possible, it's just a tradeoff to arbitrate each time ;-)

db/seeds.rb
```diff
@@ -5,3 +5,15 @@
...
 #   Character.create(name: "Luke", movie: movies.first)
+
+if Doorkeeper::Application.count.zero?
+  Doorkeeper::Application.create!(name: "Web Client", redirect_uri: "", scopes: "")
+  Doorkeeper::Application.create!(name: "iOS Client", redirect_uri: "", scopes: "")
+  Doorkeeper::Application.create!(name: "Android Client", redirect_uri: "", scopes: "")
+end
+
+User.first_or_create(
+  email: "test@nimblehq.co",
```
malparty:
```suggestion
User.find_or_create_by(email: 'test@nimblehq.co')
```

In case your DB already contains a user with another email, you might not understand why your API can't authenticate even right after generating all the seeds successfully.

With this syntax, if will check if the user with the `test@nimblehq.co` email already exists or not. If not, it will create it ;-)

db/seeds.rb
```diff
@@ -5,3 +5,15 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+
+if Doorkeeper::Application.count.zero?
+  Doorkeeper::Application.create!(name: "Web Client", redirect_uri: "", scopes: "")
+  Doorkeeper::Application.create!(name: "iOS Client", redirect_uri: "", scopes: "")
+  Doorkeeper::Application.create!(name: "Android Client", redirect_uri: "", scopes: "")
```
malparty:
```suggestion
  Doorkeeper::Application.create!(name: 'Web Client', redirect_uri: '', scopes: '')
  Doorkeeper::Application.create!(name: 'iOS Client', redirect_uri: '', scopes: '')
  Doorkeeper::Application.create!(name: 'Android Client', redirect_uri: '', scopes: '')
```

Let's use `single quote` whenever possible 🙌

malparty (changes requested):

.gitignore
phongvhd93: It is generated when I type `bundle install`

app/forms/registration_form.rb
phongvhd93: This was solved in #330251b

Gemfile
phongvhd93: I removed this gem in 3301251b because we already had `byebug`

spec/forms/registration_form_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+        password: '123456',
+        password_confirmation: '123456',
+        client_id: application.uid
+      }
+      it 'returns true' do
+        form = described_class.new
+
+        expect(form.save(params)).to be(true)
+      end
+    end
```
byhbt: The fixtures should be located within the `it ...` like the example below 

```suggestion
RSpec.describe RegistrationForm do
  describe '#save' do
    context 'given valid user input' do
      it 'returns true' do
        params = {
          email: "test+#{Time.now.to_i}@nimblehq.co",
          password: '123456',
          password_confirmation: '123456',
          client_id: application.uid
        }

        form = described_class.new

        expect(form.save(params)).to be(true)
      end
    end
```

app/forms/registration_form.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'byebug'
+
+class RegistrationForm < ApplicationForm
+  def save(create_params)
+    return unless validate_client(create_params[:client_id])
```
byhbt: Because in the controller we have `if registration_form.save(create_params)`, therefore in this FormObject we should return true/false for consistency check from the controller. 

Please check other return below as well 

```suggestion
    return false unless validate_client(create_params[:client_id])
```

app/forms/registration_form.rb
```diff
@@ -0,0 +1,32 @@
+# frozen_string_literal: true
+
+require 'byebug'
```
byhbt:
```suggestion
```
And on this file as well: `spec/forms/registration_form_spec.rb`

Let's remove debug on the ready PR. 🎨 If not, you can convert the PR to `draft` state. 😉

spec/forms/registration_form_spec.rb
```diff
@@ -0,0 +1,92 @@
...
+      user = Fabricate(:user)
+
+      params = {
+        email: user.email,
+        password: '123456',
+        password_confirmation: '123456',
+        client_id: application.uid
+      }
+
+      it 'return nil' do
```
byhbt:
```suggestion
      it 'returns nil' do
```

Please check other places as well.

.gitignore
malparty: ok, maybe from one of the gem specific to auth that I haven't used yet 🤔 
Thanks ~

app/forms/registration_form.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+class RegistrationForm < ApplicationForm
+  def save(create_params)
+    return false unless validate_client(create_params[:client_id])
+
+    user = User.new(create_params.except(:client_id))
+
+    unless user.valid?
```
malparty:
```suggestion
    if user.invalid?
```

Rails rails rails xD

app/forms/registration_form.rb
```diff
@@ -0,0 +1,30 @@
...
+      user_error = user.errors
+      errors.add(:user, user_error.full_messages.last) unless user_error.empty?
```
malparty:
```suggestion
      user_errors = user.errors
      errors.add(:user, user_errors.full_messages.last) unless user_errors.empty?
```

Also, do not feel obliged to solve all the warnings from `reek` (e.g. `calling user.errors twice`). These warnings make sense in case `.errors` was a resource heavy method (e.g. heavy calculations/processing without caching) — but `errors` is just a hash so we can call `user.errors.XXX` without any performance issue — or at least the perf difference is negligeable.

config/initializers/devise.rb
```diff
@@ -183,7 +183,7 @@
   # Email regex used to validate email formats. It simply asserts that
   # one (and only one) @ exists in the given string. This is mainly
   # to give user feedback and not to assert the e-mail validity.
-  config.email_regexp = /\A[^@\s]+@[^@\s]+\z/
+  config.email_regexp = /\A[\w+\-.]+@[a-z\d\-]+(\.[a-z]+)*\.[a-z]+\z/i
```
malparty, line 186: Should we add a test for it? 🤔 

Another possibility is to add a higher level test in `spec/forms/registration_form_spec.rb` with something like `given an invalid email`/`it has the error message` ;-)

malparty (changes requested):

app/forms/registration_form.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+class RegistrationForm < ApplicationForm
+  def save(create_params)
+    return false unless validate_client(create_params[:client_id])
+
+    user = User.new(create_params.except(:client_id))
+
+    if user.invalid?
+      errors.add(:user, user.errors.full_messages.last) unless user.errors.empty?
```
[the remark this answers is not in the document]
blyscuit: You can declare `user.errors` to solve this error.
```suggestion
      user_errors = user.errors
      errors.add(:user, user_errors.full_messages.last) unless user_errors.empty?
```

blyscuit (commented):
You might need to rebase to make the PR easier to review 🙌

app/forms/registration_form.rb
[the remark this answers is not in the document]
malparty: As said in an earlier comment, 
- it's not an `error`, just a warning to call for our attention
- ONLY fix it if calling `user.errors` is a resource-intensive method. In this case, it only points to an instance of an object, so creating a local variable won't change much (but might make code more verbose / longer to read) ;-)

app/forms/registration_form.rb
```diff
@@ -0,0 +1,30 @@
...
+
+  def validate_client(client_id)
+    unless Doorkeeper::Application.find_by(uid: client_id)
+      errors.add(:client_id, I18n.t('doorkeeper.errors.messages.invalid_client'))
+
+      return
+    end
+
+    true
+  end
```
malparty: this method returns `true` or `nil`. 
In general, I recommend to return `true/false` or `object_instance/nil`, but not returning a nullable boolean — even if "it works" ^^

Besides, some minor recommendations:
- we can have a more Rails flavor `client_valid?` naming
- we can leverage an `early return` to avoid using `unless` and make the code feel lighter 🪶 
- As we do not use the returned value of `find_by`, it will be more performant to use the [exists?](https://api.rubyonrails.org/classes/ActiveRecord/FinderMethods.html#method-i-exists-3F) method :) 

```suggestion
  def client_valid?(client_id)
    return true if Doorkeeper::Application.exists?(uid: client_id)
    
    errors.add(:client_id, I18n.t('doorkeeper.errors.messages.invalid_client'))
    
    false
  end
```

db/seeds.rb
```diff
@@ -5,3 +5,11 @@
 #
 #   movies = Movie.create([{ name: "Star Wars" }, { name: "Lord of the Rings" }])
 #   Character.create(name: "Luke", movie: movies.first)
+
+if Doorkeeper::Application.count.zero?
+  Doorkeeper::Application.create!(name: 'Web Client', redirect_uri: '', scopes: '')
+  Doorkeeper::Application.create!(name: 'iOS Client', redirect_uri: '', scopes: '')
+  Doorkeeper::Application.create!(name: 'Android Client', redirect_uri: "", scopes: '')
+end
```
malparty, line 13: Optional suggestion: Should we also use `find_or_create_by` for the `DoorKeeper::Applications` seeds?

spec/forms/registration_form_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+
+require 'rails_helper'
+
+RSpec.describe RegistrationForm do
+  describe '#save' do
+    context 'given valid user input' do
+      it 'returns true' do
+        form = described_class.new
+
+        expect(form.save(valid_signup_params)).to be(true)
```
malparty: In general, we prefer explicit tests so that we do not need to open/close multiple files when reading a test.

This means we would copy this inside every single test, and adjust the values based on the context of the test:

```
valid_signup_params = Fabricate(:application,
                                email: "test+#{Time.now.to_i}@nimblehq.co",
                                password: '123456',
                                password_confirmation: '123456',
                                client_id: application.uid)
```

It is related to this section: https://nimblehq.co/compass/development/code-conventions/ruby/rspec/#fixtures-isolation

Of course, we still need helper methods for the most common actions like `login_as user` for requests tests ;-)

spec/rails_helper.rb
```diff
@@ -11,4 +12,10 @@
 
 RSpec.configure do |config|
   config.include OAuthHelpers
+  config.include RegistrationsHelpers
+
+  config.around(:each) do |example|
+    DatabaseCleaner.strategy = example.metadata[:type] == :feature ? :seeded : :transaction
```
malparty: Do you have a use case where we need the `:seeded` strategy? 👀 

If not, I recommend to only keep `:transaction` (as using seeds in a test seems a bit contrary with our conventions to only build the fixtures we need and keep them explicit)

malparty (changes requested):

spec/requests/api/v1/registrations_controller_spec.rb
```diff
@@ -0,0 +1,67 @@
...
+        email: "test+#{Time.now.to_i}@nimblehq.co",
+        password: '123456',
+        password_confirmation: '123456',
+        client_id: application.uid
+      }
+      it 'returns http status of created' do
+        post api_v1_registrations_path, params: valid_signup_params
+
+        expect(response).to have_http_status(:created)
+      end
```
byhbt: Please adjust the other tests below as well. For preparing test data, it should be included within a specific `it ...`

```suggestion
      it 'returns the created status' do
        application ||= Fabricate(:application)
        valid_signup_params = {
          email: "test+#{Time.now.to_i}@nimblehq.co",
          password: '123456',
          password_confirmation: '123456',
          client_id: application.uid
        }

        post api_v1_registrations_path, params: valid_signup_params

        expect(response).to have_http_status(:created)
      end
```

app/forms/registration_form.rb
```diff
@@ -0,0 +1,28 @@
...
+
+class RegistrationForm < ApplicationForm
+  def save(create_params)
+    return false unless client_valid?(create_params[:client_id])
+
+    user = User.new(create_params.except(:client_id))
+
+    if user.invalid?
+      user_errors = user.errors
+      errors.add(:user, user_errors.full_messages.last) unless user_errors.empty?
```
byhbt: `last` is a bit confusing here, we can use `to_sentence` instead.  And given the user is `invalid?` then i think we don't need to check the `user.errors` is `empty?` or not anymore.

```suggestion
      errors.add(:user, user.errors.full_messages.to_sentence)
```

spec/forms/registration_form_spec.rb
```diff
@@ -0,0 +1,120 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe RegistrationForm do
+  describe '#save' do
+    context 'given valid user input' do
+      it 'returns true' do
+        application ||= Fabricate(:application)
```
byhbt: We can simplify this assignment like this, please update the spec below as well. 

The `||=` is not necessary. Reference: http://www.rubyinside.com/what-rubys-double-pipe-or-equals-really-does-5488.html


```suggestion
        application = Fabricate(:application)
```

byhbt (approved):
Approve in advance with a suggestion about changing the assignment in specs `||=`

malparty (approved):

Issue #9, opened 2023-05-29
[Backend] As a user, I can upload a CSV File and have its content saved


blyscuit:
## Why

User should be able to upload a CSV File. The content of the file should have **no header** and **contains only array of string**.

## Acceptance Criteria

- Modify endpoint from #10 to allow uploading a CSV file.
- Each keyword should be saved to the database with relationship many-user<>many-keyword.

malparty:
Should we split this story into 2 smaller ones?
- As a user, I can upload a file (create the endpoint and just return the response with the count of keywords present in the file)
- As a user, when I upload a file, it is parsed and saved in the DB (focus on the DB model and the persistence of the uploaded file)


As this is among the first main feature of the IC, we can expect many file changes and so having smaller stories would also make code review easier :)

PR #41, opened 2023-06-12
[#9] [Backend] As a user, I can upload a CSV File and have its content saved

blyscuit:
#9 

## What happened 👀

Add backend for saving uploaded CSV files to user's keywords.

Multiple line changes as a result from sample CSV.

## Insight 📝

- Modify `Keyword` model for database.
- Added `CsvValidator` for validating CSV file.
- Added `CsvVUploadForm`.
- Create One<>One relationships between `Keywords` and `User`.
- Added testing CSV files via fixtures.
- Added `Inflector` for CSV.

## Proof Of Work 📹

Screen Shot 2023-06-12 at 15 34 30


app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,44 @@
...
+    errors.empty?
+  end
+
+  private
+
+  def save_keywords
+    keyword_hash.each do |hash|
+      keyword = Keyword.find_or_create_by!(hash)
+      keyword_user_hash = { keyword: keyword, user: @user }
+      KeywordUser.find_or_create_by!(keyword_user_hash)
```
malparty:
```suggestion
      @current_user.keywords << keywords
```

`<<` in Rails ActiveRecord will add the item to the collection AND save the item.
We can leverage the associations instead of instantiating a join-table class ;-)

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,44 @@
...
+
+class CSVUploadForm
+  include ActiveModel::Validations
+
+  attr_reader :file
+
+  validates_with CSVValidator
+
+  def initialize(user)
+    @user = user
```
malparty:
```suggestion
  def initialize(current_user)
    @current_user = current_user
```

I recommend sticking with `current_user` to add more context (the cases when we have different users than the current users are quite common too so this helps be sure about which user it is :) )

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,44 @@
...
+
+  def save(file)
+    @file = file
+    return false unless file && valid?
+
+    begin
+      save_keywords
+    rescue ActiveRecord::StatementInvalid
+      errors.add(:base, I18n.t('csv.validation.bad_keyword_length'))
+    end
```
malparty:
```suggestion
    save_keywords
```

I recommend handling this validation before calling `save`. In a similar way, you already have a `valid?` method and a `validator class`. You can add some checks before the saving. 

There will always be a risk of "DB error", but it is better to let the error be raised so that if you have a logging tool, you will have the full error. If we rescue the `generic` error `StatementInvalid` and assign `bad_keyword_lenght`, it might tell the user (and us) the wrong thing (maybe the error was raised due to a special char, or a corrupted CSV file, etc... be the user is informed that the keyword has the wrong length).

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,44 @@
...
+      keyword_user_hash = { keyword: keyword, user: @user }
+      KeywordUser.find_or_create_by!(keyword_user_hash)
+    end
+  end
+
+  def keyword_hash
+    CSV.read(file).filter_map { |row|
+      name = row.join(',')
+      name.blank? ? false : { name: name }
+    }
```
malparty: How do we store the keywords? 👀 

From my reading, I see that a `keyword` object will have the column `name` with a value like `keywoard1,keyword2,keyword3,keyword4,keyword5`. Is that correct? 💭 

If yes, I recommend 2 options:
- either keep the same format, but we must rename `keyword` with `keyword_batch` or `keyword_set` or `keyword_group`
- either see if we can change the structure so that 1 keyword represent only 1 keyword (instead of a list of keywords).

This is mainly to ensure the name of our variables correctly describe what they hold.

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,44 @@
...
+  def save_keywords
+    keyword_hash.each do |hash|
+      keyword = Keyword.find_or_create_by!(hash)
+      keyword_user_hash = { keyword: keyword, user: @user }
+      KeywordUser.find_or_create_by!(keyword_user_hash)
+    end
+  end
+
+  def keyword_hash
+    CSV.read(file).filter_map { |row|
```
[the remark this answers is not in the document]
malparty:
```suggestion
    CSV.read(file).filter_map do |row|
```
```
      # [...]
    end
```

app/models/keyword_user.rb
```diff
@@ -0,0 +1,6 @@
+# frozen_string_literal: true
+
+class KeywordUser < ApplicationRecord
+  belongs_to :keyword
+  belongs_to :user
```
malparty: let's use the `inverse of` as [explained in compass](https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#convention-13)

```suggestion
  belongs_to :keyword, inverse_of: :keyword_users
  belongs_to :user, inverse_of: :keyword_users
```

Same for the other models~

app/validators/csv_validator.rb
```diff
@@ -0,0 +1,36 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CSVValidator < ActiveModel::Validator
+  def validate(csv_form)
+    @csv_form = csv_form
```
malparty: The `Validator` pattern in Rails has 1 big drawback: It acts nearly as a Singleton :harold:.
See: https://api.rubyonrails.org/v7.0.5/classes/ActiveModel/Validator.html#:~:text=Note%20that%20the%20validator%20is%20initialized%20only%20once%20for%20the%20whole%20application%20life%20cycle%2C%20and%20not%20on%20each%20validation%20run.

This means that a validator will be initialized once for the whole application lifecycle — so an instance variable might leak from one request to another (which is a security issue). For that reason, we must not use instance variables in a validator 👍

app/validators/csv_validator.rb
```diff
@@ -0,0 +1,36 @@
...
+  def add_error(type)
+    csv_form.errors.add(:base, I18n.t("csv.validation.#{type}"))
+  end
+
+  def valid_count?
+    CSV.read(file).count.between?(1, 1000)
+  end
+
+  def valid_extension?
+    File.extname(file) == '.csv'
```
malparty:
```suggestion
    File.extname(file) == CSV_EXTENSION
```

we can declare it as a constant at the top of the class: `CSV_EXTENSION = '.csv'.freeze` ;-)

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,44 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CSVUploadForm
```
byhbt:
```suggestion
class CsvUploadForm
```

Follow our naming convention here: https://nimblehq.co/compass/development/code-conventions/#convention-19
The `CSV` should be `Csv`

config/locales/en.yml
```diff
@@ -31,3 +31,9 @@
 
 en:
   hello: "Hello world"
+  csv:
+    upload_success: 'Your file has been uploaded!'
+    upload_btn: 'Upload a CSV File'
```
```suggestion
    upload_btn: 'Upload a CSV File'
    upload_success: 'Your file has been uploaded!'
```

I recommend to order YML alphabetically for 2 reasons:
- it's easier to find/check a key later
- if you use `phrase` (in most client projects), we regularly need to `pull` the translations (as they get updated by PM/Client in the phrase cloud) — pulled translations will be ordered ABC, so it means less code changes when we already have the same ordering :)

db/migrate/20230609095923_keyword.rb
```diff
@@ -0,0 +1,8 @@
+class Keyword < ActiveRecord::Migration[7.0]
+  def change
+    create_table :keywords do |t|
+      t.string :name
```
malparty, line 4:
```suggestion
      t.string :name, index: true
```

As we will perform `search` on the keyboard names, it is nice to add an `index` on it.

In a simplified verison, the arbitration for indexes is:
- is update/write speed critical? if yes, avoid indexes (as each value change will require the DB to rebuild the index)
- is read/retrieval/search/ordering critical? if yes, you probably need an index

app/validators/csv_validator.rb
byhbt, line 1: We also need to write a spec for this custom validator.

malparty (changes requested):
An early review.
I stop until we answered the question in Slack about the many-to-many association :)

app/validators/csv_validator.rb
blyscuit: This is tested as part of `csv_upload_form`, do you think it's sufficient?
Screen Shot 2023-06-13 at 12 30 43

Currently I am not sure how to properly test just a validator file so I will be needing some advice 🙂

app/forms/csv_upload_form.rb
blyscuit: Thank you. I like this Ruby style naming more.

blyscuit:
@malparty I updated the database relationships and modified `csv_upload_form` to match the new change. Please help me see if there's any concerns.

app/validators/csv_validator.rb
blyscuit: I updated the PR with the validator spec. Thank you for the advice.

app/validators/csv_validator.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+require 'csv'
+
+class CsvValidator < ActiveModel::Validator
+  CSV_EXTENSION = '.csv'
+
+  def validate(csv_form)
+    validate_file(csv_form)
```
malparty:
```suggestion
    file = csv_form.file
    add_error(csv_form, :wrong_count) unless valid_count(file)
    add_error(csv_form, :wrong_type) unless valid_extension(file)
```

It seems like we do not need an additional method `validate_file` ;-) 
(if needed later we can bring it back)

malparty:
@blyscuit the DB model looks good now, but there are still some other unresolved comments like the `inverse_of` and [this one](https://github.com/nimblehq/ic-rails-phong-bliss/pull/41#discussion_r1227449671)

app/forms/csv_upload_form.rb
blyscuit: The keyword is saved individually, the join logic here is to connects keyword in the format `hello, world` > `['hello', 'world']` > `'hello, world'`.

The default of `CSV.read` is that one row = array of strings separated by `,`. So we have to join them.

The idea of the DB is `1 keyword represent only 1 keyword`.

app/models/keyword_user.rb
blyscuit: I removed this file and add `inverse_of` to `keywords.rb` 🙏

app/forms/csv_upload_form.rb
```diff
@@ -0,0 +1,35 @@
...
+    errors.empty?
+  end
+
+  private
+
+  def keyword_hash
+    CSV.read(file).filter_map do |row|
+      name = row.join(',')
+      name.blank? ? false : { name: name, user_id: @current_user.id }
+    end
```
byhbt: Can we try this one, i think it a bit cleaner: 🎨 

```suggestion
    CSV.read(file).filter_map do |row|
      name = row.join(',')
      next if name.blank?
     { name: name, user_id: current_user.id }
    end
```

app/validators/csv_validator.rb
```diff
@@ -0,0 +1,27 @@
...
+  end
+
+  private
+
+  def add_error(csv_form, type)
+    csv_form.errors.add(:base, I18n.t("csv.validation.#{type}"))
+  end
+
+  def valid_count(file)
+    CSV.read(file).count.between?(1, 1000)
```
byhbt: Avoiding the magic number. 💡 

```suggestion
    # Let define a 1000 as the constant under CSV_EXTENSION

    CSV.read(file).count.between?(1, LIMIT_ROWS)
```

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        expect(form.errors.full_messages).to include(I18n.t('csv.validation.wrong_type'))
+      end
+    end
+
+    context 'given a file with 9 keywords and 1 blank' do
+      it 'saves 9 keywords with the correct user' do
+        user = Fabricate(:user)
+        form = described_class.new(user)
+        form.save(file_fixture('csv/blank.csv'))
+        expect(keyword_for_user(user).count).to eq(9)
```
byhbt, line 69: Just make it consistent for the expect, add one empty line before it to improve the readibility. 

```suggestion

        expect(keyword_for_user(user).count).to eq(9)
```

spec/forms/csv_upload_form_spec.rb
```diff
@@ -0,0 +1,74 @@
...
+        form = described_class.new(user)
+        saved = form.save(file_fixture('csv/valid.csv'))
+        expect(saved).to be(true)
+      end
+
+      it 'saves 10 keywords with the correct user' do
+        user = Fabricate(:user)
+        form = described_class.new(user)
+        form.save(file_fixture('csv/valid.csv'))
+        expect(keyword_for_user(user).count).to eq(10)
```
byhbt, line 21: Usually when we try to assert the count change or not change after an insert, we can use this syntax in Rspec. 

```suggestion
expect do
  form.save(file_fixture('csv/valid.csv'))
end.to change(Keyword, :count).by(10)
```

spec/validators/csv_validator_spec.rb
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  context 'given a valid file with blank keywords' do
+    it 'validates to valid' do
+      form = CsvValidatorHelper.new(file_fixture('csv/blank.csv'))
+      expect(form.valid?).to be(true)
+    end
+  end
+
+  context 'given a file with 1001 keywords' do
```
byhbt: If we defined a constant, we can update the context here. The goal is to avoid the magic number, in case it changes, then we no worry about updating the related places like this. 

```suggestion
  context 'given a file with the number of keywords greater than the LIMIT' do
```

app/forms/csv_upload_form.rb
blyscuit: @byhbt Somehow this `filter_map` requires explicit `false` or `nil` which is why I have to include `if else` here, I also tried with `return unless` etc. but the `filter_map` didn't behave correctly.

app/validators/csv_validator.rb
blyscuit: Sounds good. I updated.

spec/forms/csv_upload_form_spec.rb
blyscuit: I think the assertion here is different, where `change(Keyword, :count).by(10)` checks for the 10 addition but `expect(keyword_for_user(user).count).to eq(10)` checks if the  keywords belongs to the correct user.

byhbt: Ah I got your point here now. Thanks!

byhbt (approved):

app/models/keyword.rb
```diff
@@ -1,7 +1,6 @@
 # frozen_string_literal: true
 
-class Keyword
-  include ActiveModel::API
-
-  attr_accessor :id, :name
+class Keyword < ApplicationRecord
+  validates :name, presence: true
+  belongs_to :user, inverse_of: :keywords
```
malparty, line 5: The user model is missing the association. So if we manipulate a keyword instance, we will get:

```
ActiveRecord::InverseOfAssociationNotFoundError: Could not find the inverse association for user (:keywords in User)
from /Users/xavier/.rvm/gems/ruby-3.0.1@google_search_ruby/gems/activerecord-7.0.1/lib/active_record/reflection.rb:236:in `check_validity_of_inverse!'
```

app/forms/csv_upload_form.rb
malparty: To be honest, I did not understand the code until I debug it and run a test to see what `row` contained 🥲

As a first suggesiton, should we rename `row` to be more explicit that it is an array of columns for 1 row?
Then we can see the [CSV::Row documentation](https://ruby-doc.org/stdlib-3.0.1/libdoc/csv/rdoc/CSV/Row.html) to see if there is another way than `join` — so that if we have a CSV using `;` as the separator char, it would still work. Finally we need to `chomp` to remove the final `\n` from the `to_csv` method 😇 

```suggestion
    CSV.read(file).filter_map do |row_columns|
      name = row_columns.to_csv.chomp
      name.blank? ? false : { name: name, user_id: @current_user.id }
    end
```


Also, as `insert_all` is skipping validations, should we build and validate a keyword object at this stage? 🤔 

That would give us:

```
  def keywords
    CSV.read(file).filter_map do |row_columns|
      keyword = Keyword.new(name: row_columns.to_csv.chomp, user_id: @current_user.id)

      keyword if keyword.valid?
    end
  end
```

and so the save method can be simplified to just save keywords (a little bit less performant*** but will still run a single DB query):

> _*** the performance reduction is actually due to the instantiation and validation made inside the `keywords` method_

```
  def save(file)
    @file = file
    return false unless file && valid?

    ActiveRecord::Base.transaction { keywords.each(&:save) }

    errors.empty?
  end
```

The tests are still passing with this code suggestion so I think it's OK.

malparty (changes requested):
We are getting there! Code structure looks good, tests as well. I was able to focus on the core business (which I have some last suggestions 😇 )

app/forms/csv_upload_form.rb
blyscuit: I compared `single save` and `import_all`, the performance seems similar with low entries count <1000. 👍

malparty (approved):
🚀

app/forms/csv_upload_form.rb
malparty: My bad, at the time I wrote this I though csv would be a single line of multiple keywords (instead of 1 keyword per line) 😇

phongvhd93 (approved):

PR #42, opened 2023-06-13
[#15] [Backend] As a user, I can upload a list of keywords to trigger a Google search

### LINKED ISSUE #15: [Backend] As a user, I can upload a list of keywords to trigger a Google search
## Why

## Acceptance Criteria

- Add backend function for accessing Google search page with the Keywords
- This function should be called when user is uploading a CSV file.

blyscuit:
- close #15 

## What happened 👀

Add backend logic for parsing CSV files to HTML request, then parsing to search results.

Search result includes:

## Insight 📝

- Added `Google::ParserService` for parsing HTML to Google keyword result.
- Added `WebpageFetchService` for fetching any webpage.
- Added `Google::KeywordService` for abstracting constants with Google calls.
- Services call will be done in #16 along with the test for controller.

## Proof Of Work 📹

Using the call to get live data for keyword `vpn`
```
      def index
        result = Google::KeywordService.new('vpn').call()
        render json: result
      end
```
Screen Shot 2023-06-19 at 17 51 38


app/services/google/keyword_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Google
+  class KeywordService
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) '\
+                 'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    BASE_SEARCH_URL = 'https://www.google.com/search'
+
+    def initialize(keyword, lang: 'en')
```
malparty:
```suggestion
    def initialize(keyword, lang: :en)
```

Whenever possible, using symbol is better for performance (and shorter to write ^^)

app/services/google/keyword_service.rb
```diff
@@ -0,0 +1,20 @@
+# frozen_string_literal: true
+
+module Google
+  class KeywordService
```
malparty:
```suggestion
  class SearchService
```
In general, [services](https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#services) start with a verb to describe the action that will be triggered when calling the `call` method.

As this service can be used to search anything on Google (not just a keyword but also a whole phrase), I think we do not need to add the word "keyword" in its name.

That would give us: `Google::SearchService`. :)

app/services/google/keyword_service.rb
```diff
@@ -0,0 +1,20 @@
...
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) '\
+                 'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    BASE_SEARCH_URL = 'https://www.google.com/search'
+
+    def initialize(keyword, lang: 'en')
+      @escaped_keyword = CGI.escape(keyword)
+      @url = "#{BASE_SEARCH_URL}?q=#{@escaped_keyword}&hl=#{lang}&gl=#{lang}"
```
malparty: (The bellow is a totally optional and opinionated suggestion)

I generally prefer performing the least in the initialize method. This allows to delay some calls (e.g. formatting the URL) and to only call them when we are sure we will consume their output.

This would look like:

```
    def initialze(keyword, lang: :en)
      @keyword = keyword
      @lang = lang
    end

    def call
      html = WebpageFetchService.new(url, USER_AGENT).call

      ParserService.new(html.body).call if html
    end

    private

    def url
      escaped_keyword = CGI.escape(keyword)

      "#{BASE_SEARCH_URL}?q=#{@escaped_keyword}&hl=#{lang}&gl=#{lang}"
    end
```

In a simple service like this one, it does not change much. But when the class becomes more complex, this can make a big difference ;)

app/services/google/keyword_service.rb
```diff
@@ -0,0 +1,20 @@
...
+
+module Google
+  class KeywordService
+    USER_AGENT = 'Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) '\
+                 'AppleWebKit/537.36 (KHTML, like Gecko) Chrome/91.0.4472.77 Safari/537.36'
+
+    BASE_SEARCH_URL = 'https://www.google.com/search'
+
+    def initialize(keyword, lang: 'en')
+      @escaped_keyword = CGI.escape(keyword)
```
malparty:
```suggestion
      escaped_keyword = CGI.escape(keyword)
```

This one does not need to be assigned as instance variable ;-)

app/services/google/parser_service.rb
```diff
@@ -0,0 +1,62 @@
+# frozen_string_literal: true
+
+module Google
+  class ParserService
```
malparty:
```suggestion
  class ParseResultsService
```

Same idea for using an action verb ;-)

app/services/google/parser_service.rb
```diff
@@ -0,0 +1,62 @@
...
+    TOP_ADS_SECTION_ID = 'tvcap'
+    BOTTOM_ADS_SECTION_ID = 'bottomads'
+    ADS_CLASS = 'uEierd'
+    SEARCH_SECTION_ID = 'search'
+    SEARCH_RESULT_CLASS = 'yuRUbf'
+
+    def initialize(html)
+      @html = html
+      @doc = Nokogiri::HTML(html)
```
malparty: I would recommend also to move the Nokogiri call into the call method - to delay its initialization until the time we are sure it will be needed :)

app/services/google/parser_service.rb
```diff
@@ -0,0 +1,62 @@
...
+    private
+
+    attr_reader :html, :doc
+
+    def ads_top_count
+      doc.css("div[id='#{TOP_ADS_SECTION_ID}']").css("div[class='#{ADS_CLASS}']").count
+    end
+
+    def ads_page_count
+      doc.css("div[class='#{ADS_CLASS}']").count
```
malparty:
```suggestion
      @ads_page_count ||= doc.css("div[class='#{ADS_CLASS}']").count
```

As we are calling this method twice, it is interesting to cache the value.

Same for the `result_count` method ;-)

app/services/webpage_fetch_service.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class WebpageFetchService
```
malparty:
```suggestion
class FetchWebpageService
```

spec/services/google/keyword_service_spec.rb
```diff
@@ -0,0 +1,68 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::KeywordService, type: :service do
+  describe '#call' do
+    context 'given a keyword with valid result' do
```
malparty:
```suggestion
    context 'given a keyword with valid results' do
```

spec/services/google/keyword_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+      it 'returns 3 top ads count' do
```
malparty:
```suggestion
      it 'returns 3 as top ads count' do
```

or `it has a top ads count of 3`. 
Otherwise one might understand there are 3 different values returned as top ads count.

spec/services/google/keyword_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        expect(result[:ads_page_count]).to eq(4)
+      end
+
+      it 'returns the correct ads urls' do
+        stub_request(:get, 'https://www.google.com/search?gl=en&hl=en&q=vpn').to_return(body: file_fixture('html/valid_google.html').read)
+        result = described_class.new('vpn').call
+        expect(result[:ads_top_urls]).to contain_exactly('https://hubstaff.com/features/automatic-time-tracker', 'https://hubstaff.com/time_tracking_software', 'https://hubstaff.com/pricing', 'https://hubstaff.com/demo', 'https://www.odoo.com/app/timesheet', 'https://zapier.com/apps/timely-time-tracking/integrations')
+      end
+
+      it 'returns 10 results count' do
```
malparty:
```suggestion
      it 'counts 10 results' do
```
An alternative that works too 😇

spec/services/google/keyword_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        expect(result[:result_urls]).to contain_exactly('https://www.forbes.com/advisor/business/software/free-time-tracking-apps/', 'https://zapier.com/blog/best-time-tracking-apps/', 'https://clockify.me/', 'https://myhours.com/best-time-tracking-apps', 'https://toggl.com/', 'https://toggl.com/blog/best-time-tracking-apps', 'https://desktime.com/best-employee-time-tracking-software', 'https://play.google.com/store/apps/details?id=com.aloggers.atimeloggerapp&hl=en&gl=US', 'https://trackingtime.co/', 'https://collegeinfogeek.com/time-tracking-app/')
+      end
+
+      it 'returns 14 total links count' do
+        stub_request(:get, 'https://www.google.com/search?gl=en&hl=en&q=vpn').to_return(body: file_fixture('html/valid_google.html').read)
+        result = described_class.new('vpn').call
+        expect(result[:total_link_count]).to eq(14)
+      end
+
+      it 'returns correct html' do
```
malparty:
```suggestion
      it 'returns the html of the initial request' do
```

Whenever possible, let's not use "correct" to describe the tests, because the test description should explain what to expect from the method. All methods are expected to return the correct data, but that doest not explain what is returned 😇

spec/services/google/keyword_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+        expect(result[:ads_top_urls]).to contain_exactly('https://hubstaff.com/features/automatic-time-tracker', 'https://hubstaff.com/time_tracking_software', 'https://hubstaff.com/pricing', 'https://hubstaff.com/demo', 'https://www.odoo.com/app/timesheet', 'https://zapier.com/apps/timely-time-tracking/integrations')
+      end
+
+      it 'returns 10 results count' do
+        stub_request(:get, 'https://www.google.com/search?gl=en&hl=en&q=vpn').to_return(body: file_fixture('html/valid_google.html').read)
+        result = described_class.new('vpn').call
+        expect(result[:result_count]).to eq(10)
+      end
+
+      it 'returns the correct result urls' do
```
malparty:
```suggestion
      it 'returns the url of the results that are not ads' do
```

spec/services/google/keyword_service_spec.rb
```diff
@@ -0,0 +1,68 @@
...
+
+    context 'given a keyword with 500 status error result' do
+      it 'returns nil' do
+        stub_request(:get, 'https://www.google.com/search?gl=en&hl=en&q=vpn').to_return(status: 500)
+        result = described_class.new('vpn').call
+        expect(result).to be_nil
+      end
+    end
+
+    context 'given a keyword with an invalid html result' do
```
malparty:
```suggestion
    context 'given the request returned an invalid html result' do
```

In this test, we use the keyword `vpn`, but the keyword `vpn` is not the reason why we get an invalid HTML result. The reason is that the request had an issue (wether it's on Google side or on the app side).

spec/services/google/parser_service_spec.rb
```diff
@@ -0,0 +1,45 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::ParserService, type: :service do
+  describe '#call' do
+    context 'given a valid google page' do
```
malparty:
```suggestion
    context 'given a valid Google page' do
```

app/services/webpage_fetch_service.rb
```diff
@@ -0,0 +1,24 @@
...
+  end
+
+  def call
+    result = HTTParty.get(@uri, { headers: { 'User-Agent' => @user_agent } })
+
+    return false unless valid_result? result
+
+    result
+  rescue HTTParty::Error, Timeout::Error, SocketError
+    false
```
malparty:
```suggestion
    return unless valid_result? result

    result
  rescue HTTParty::Error, Timeout::Error, SocketError
    nil
```

Let's not return `result || nil`: Either we have a boolean method that returns `true || false`, either we have an object method that results `object || nil` ;-)

malparty: (tests will need adjustment accordingly ;-) )

spec/services/webpage_fetch_service_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+    context 'when querying a webpage with User-Agent' do
+      it 'returns an HTTParty Response' do
+        url = FFaker::Internet.http_url
+        user_agent = FFaker::FreedomIpsum.word
+        WebMock.stub_request(:get, url).with(headers: { 'User-Agent' => user_agent })
+        result = described_class.new(url, user_agent).call
+
+        expect(result).to be_an_instance_of(HTTParty::Response)
+      end
+    end
```
malparty: When reading these two context, it's not clear what is the difference between using a User Agent and not using a user agent. 

This gives us 2 options:
- Either there is no much difference and then only 1 test is needed
- Either we are not asserting enough things — Is there the user agent in the headers of the response object? If yes we might want to test that // or maybe the returned HTML is difference 🤔 

I do not have the immediate solution but my point is just: Think of tests are a documentation, each context should clearly explain why it is different from the other contexts — i.e. why should I use `user_agent`, when is it ok to not provide the user agent, etc...

malparty (changes requested):

spec/services/google/keyword_service_spec.rb
blyscuit: What do you think about `returns results count of 10` so that we can keep the `returns {value}` format 🙏

spec/services/webpage_fetch_service_spec.rb
blyscuit: I moved the `user-agent` out of this class as it does not belong here, added `headers` instead.

spec/services/google/keyword_service_spec.rb
malparty: Perfectly fine as well! :)

app/services/google/parse_result_service.rb
```diff
@@ -0,0 +1,62 @@
...
+    SEARCH_SECTION_ID = 'search'
+    SEARCH_RESULT_CLASS = 'yuRUbf'
+
+    def initialize(html)
+      @html = html
+    end
+
+    def call
+      @doc = Nokogiri::HTML(html)
+      {
```
malparty, line 18:
```suggestion

      {
```

Let's just use an empty line before the "return" statement

spec/services/fetch_webpage_service_spec.rb
```diff
@@ -0,0 +1,38 @@
...
+        url = FFaker::Internet.http_url
+        stub_request(:get, url).with(headers: headers)
+        result = described_class.new(url, headers).call
+
+        expect(result).to be_an_instance_of(HTTParty::Response)
+      end
+    end
+
+    context 'when querying a webpage returns 500 error' do
+      it 'returns false' do
```
malparty:
```suggestion
      it 'returns nil' do
```

malparty (approved):
Approved with 2 last minor comments :)

byhbt (approved):

PR #43, opened 2023-06-14
[#10] [API] As a user, I can upload a CSV File and have its content saved

### LINKED ISSUE #10: [API] As a user, I can upload a CSV File and have its content saved
## Why

User should be able to upload a CSV File. The content of the file should have **no header** and **contains only array of string**.

## Acceptance Criteria

- Create **authenticated** endpoint `POST` `/api/v1/keyword`
    - Input:
        - Files: `CSV`
        - `Authentication` Header: `Bearer {Token}`
    - Returns `200` and empty JSON `{meta: message}` if success.
    - Returns error `400` if CSV is missing, non-CSV was uploaded, or CSV is in the wrong format.

blyscuit:
- close #10

## What happened 👀

Create **authenticated** endpoint `POST` `/api/v1/keyword`
    - Input:
        - Files: `CSV`
        - `Authentication` Header: `Bearer {Token}`
    - Returns `200` and empty JSON `{meta: message}` if success.
    - Returns error `400` if CSV is missing, non-CSV was uploaded, or CSV is in the wrong format.
    
## Insight 📝

- Add `create_token_header` to `o_auth_helper`.
- Add `ErrorHandlerConcern` for quick error render for API.
- `keywords_controller` `#post` uses existing `csv_form`.

## Proof Of Work 📹

Screen Shot 2023-06-19 at 11 28 4
<img width=
6" src="https://github.com/nimblehq/ic-rails-phong-bliss/assets/6356137/b46db5c2-224e-4561-b229-3597658ad69e">

### DIFF
--- a/app/controllers/concerns/api/v1/error_handler_concern.rb
+++ b/app/controllers/concerns/api/v1/error_handler_concern.rb
@@ -1,0 +1,20 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    module ErrorHandlerConcern
+      extend ActiveSupport::Concern
+
+      private
+
+      def render_error(detail, source: nil, status: :unprocessable_entity)
+        error = build_error(detail: detail, source: source)
+        render_errors [error], status
+      end
+
+      def render_errors(jsonapi_errors, status = :unprocessable_entity)
+        render json: { errors: jsonapi_errors }, status: status
+      end
+    end
+  end
+end
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,7 +32,7 @@
 en:
   hello: "Hello world"
   csv:
-    upload_success: 'Your file has been uploaded!'
+    upload_success: 'Your file has been uploaded successfuly!'
     validation:
       wrong_count: 'The csv file should contain between 1 and 1000 keywords'
       wrong_type: 'Please ensure the provided file type is csv'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -32,6 +32,7 @@
 en:
   hello: "Hello world"
   csv:
+    upload_success: 'Your file has been uploaded!'
     validation:
       wrong_count: 'The csv file should contain between 1 and 1000 keywords'
       wrong_type: 'Please ensure the provided file type is csv'
--- a/config/locales/en.yml
+++ b/config/locales/en.yml
@@ -34,5 +34,5 @@ en:
   csv:
     upload_success: 'Your file has been uploaded successfuly!'
     validation:
-      wrong_count: 'The csv file should contain between 1 and 1000 keywords'
-      wrong_type: 'Please ensure the provided file type is csv'
+      wrong_count: 'The CSV file should contain between 1 and 1000 keywords'
+      wrong_type: 'Please ensure the provided file type is CSV'
--- a/spec/requests/api/v1/keywords_controller_spec.rb
+++ b/spec/requests/api/v1/keywords_controller_spec.rb
@@ -3,17 +3,13 @@
 RSpec.describe Api::V1::KeywordsController, type: :request do
   describe 'GET#index' do
     it 'returns success status' do
-      header, = create_token_header
-
-      get api_v1_keywords_path, headers: header
+      get api_v1_keywords_path, headers: create_token_header
 
       expect(response).to have_http_status(:success)
     end
 
     it 'returns the valid keywords' do
-      header, = create_token_header
-
-      get api_v1_keywords_path, headers: header
+      get api_v1_keywords_path, headers: create_token_header
 
       keywords = JSON.parse(response.body, symbolize_names: true)[:data]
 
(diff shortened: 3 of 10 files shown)


malparty:
Let's not forget the PR metadata before marking it as ready for review 😇 

image

config/locales/en.yml
```diff
@@ -32,6 +32,7 @@
 en:
   hello: "Hello world"
   csv:
+    upload_success: 'Your file has been uploaded!'
```
```suggestion
    upload_success: 'Your file has been uploaded successfuly!'
```

Optional suggestion :)

spec/support/oauth_helpers.rb
```diff
@@ -35,4 +35,13 @@ def token_refresh_params(refresh_token)
...
   end
+
+  def create_token_header(user = nil)
+    user ||= Fabricate(:user)
+
+    application = Fabricate(:application)
+    access_token = Fabricate(:access_token, resource_owner_id: user.id, application_id: application.id)
+
+    [{ 'Authorization' => "Bearer #{access_token.token}" }, user]
+  end
```
malparty: I'm not sure if my suggestion works, but it would be nice if we could use this helper like `login_as user` instead of creating the token and parring them manually to the request each time 🤔 

```suggestion
  def login_as(user = nil)
    user ||= Fabricate(:user)

    application = Fabricate(:application)
    access_token = Fabricate(:access_token, resource_owner_id: user.id, application_id: application.id)

    request.headers['Authorization'] = "Bearer #{access_token.token}"
    
    user
  end
```

I let you try~ 🙌

app/controllers/concerns/api/v1/error_handler_concern.rb
malparty, line 1: This module doesn't need to extend the concern helpers as we do not use any feature from it (e.g. the `included do [...] end` block to add Rails-specific DSL or callbacks). 

Also, the naming convention for concerns (or enabling modules) is to end with `able`. For example `ErrorRenderable` (able to render errors). 

So I'd recommend to rename it ErrorRenderable and to remove the "extend ActiveSupport::Concern". That said IMO we can keep it in the `concerns` folder (as even without using 'ActiveSupport::Concern', it still acts as a concern 😇 )

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -3,18 +3,78 @@
...
+
+    context 'when user is not signed in' do
+      it 'returns unauthorized status' do
+        params = { 'file' => fixture_file_upload('csv/valid.csv') }
+        post api_v1_keywords_path, params: params
+
+        expect(response).to have_http_status(:unauthorized)
+      end
+    end
+  end
```
malparty, line 83: Should we also test the error messages? 👀

malparty (changes requested):

phongvhd93:
It would be best if you included a screenshot of the CSV file's format. So we can test the PR easier

spec/support/oauth_helpers.rb
blyscuit: @malparty I also would like this to work as well but from limited understanding, `request` is not accessible at this point for this version of RSpec? Not sure is there any import I am missing.

blyscuit: From my understanding, the Request spec cannot get access to the request and request.header, we have to use [Controller spec](https://rspec.info/features/6-0/rspec-rails/controller-specs/) instead. But the Controller spec does not seem to be made for API testing like we are doing. Also I found this section

> We encourage you to use [request specs](https://rspec.info/features/6-0/rspec-rails/controller-specs/request-specs/request-spec) if you want to set headers in your call. If you still want to use controller specs with custom http headers you can use request.headers:

spec/requests/api/v1/keywords_controller_spec.rb
blyscuit: Added error messages tests. 👍

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -3,18 +3,93 @@
 RSpec.describe Api::V1::KeywordsController, type: :request do
   describe 'GET#index' do
     it 'returns success status' do
-      get api_v1_keywords_path
+      header, = create_token_header
```
byhbt: just small suggestion, 🎨  we can define a variable with `_` underscore to mark it as unused, and it makes the pattern matching become clearer.

```suggestion
      header, _user = create_token_header
```

spec/support/oauth_helpers.rb
malparty: Oh, I see. 
The `request` variable is actually present and reachable. The problem is that the variable is `nil` because the test did not perform any request yet.

There are two options then:
- A) either keep the current implementation 
- B) either replace the token fabrication with a call to the login API so that it initialize the request object and provide you in the response the required token

Both look fine. A) is probably farter to execute // B) looks a bit more standard 😇

app/controllers/concerns/.keep
malparty, line 1: I guess we can remove this file?

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -3,18 +3,93 @@
...
 
       keywords = JSON.parse(response.body, symbolize_names: true)[:data]
 
       expect(keywords[0][:attributes][:name]).to eq('First keyword')
       expect(keywords[1][:attributes][:name]).to eq('Second keyword')
     end
   end
+
+  describe 'POST#index' do
+    context 'when csv file is valid' do
```
```suggestion
    context 'when CSV file is valid' do
```

same for the many other locations Comma Separated Values is used.

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -3,18 +3,93 @@
...
+    context 'when csv file is in the wrong type' do
+      it 'returns unprocessable_entity status' do
+        header, = create_token_header
+        params = { 'file' => fixture_file_upload('csv/wrong_type.txt') }
+        post api_v1_keywords_path, params: params, headers: header
+
+        expect(response).to have_http_status(:unprocessable_entity)
+      end
+
+      it 'returns a wrong_type error' do
```
```suggestion
      it 'returns the wrong_type error' do
```

Because there is only 1 "wrong type error".

malparty (approved):
Approved with a few last minor suggestions. For the test helper method we can keep your initial approach too 🙌

spec/support/oauth_helpers.rb
blyscuit: Let me try `B`, not sure if the `request` variable from the login call and the call from the spec will be the same.

app/controllers/concerns/.keep
blyscuit: We are removing this file 😄

spec/support/oauth_helpers.rb
blyscuit: I updated the call for `create_token_header` to make it more streamline for the call-side. Since we are passing `user?` object we don't have to return it if the spec requires a user check.

I was not able to do `B` because the `request` from the first API call is not used in the following API calls.

spec/requests/api/v1/keywords_controller_spec.rb
blyscuit: I updated `create_token_header` to return only header 🙏

phongvhd93 (approved):

byhbt (approved):

spec/support/oauth_helpers.rb
malparty: That's strange. Maybe it works a bit differently in a web API (in Flexi we just pass the devise headers to the request instance and it would work for all calls. Or maybe the helper from Devise is doing additional things 🙈 ).

Let's keep your current implementation then :)

PR #51, opened 2023-06-19
[#40] [Backend] As a user, I can verify my account with the link in my account verification email

### LINKED ISSUE #40: [Backend] As a user, I can verify my account with the link in my account verification email
## Why

To allow a user confirms their email after a user taps on the link included in the account verification email.

## Acceptance Criteria
- Return with HTTP status code 200
- Update the field `is_verified` in User table to true
- Write tests

phongvhd93:
- close #40

## What happened 👀

Configure `routes` to enable email confirmation with a link feature. I also added a simple controller to display the success messages after a user confirms the email.

## Insight 📝

- Disable sending confirmation emails in `Development` environment.
- Bring `devise_for :users` block at the top level to `API > V1`.
- Configure routes for Devise's confirmation controller.
- Add a new `success_messages_controller` for displaying a simple success message after a user confirms their email.

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-phong-bliss/assets/22606906/046060d8-d8d8-4884-b7b2-5e4c9e7cacee

### DIFF
--- a/app/controllers/api/v1/confirmations_controller.rb
+++ b/app/controllers/api/v1/confirmations_controller.rb
@@ -2,6 +2,10 @@
 
 module Api
   module V1
-    class ConfirmationsController < Devise::ConfirmationsController; end
+    class ConfirmationsController < Devise::ConfirmationsController
+      def after_confirmation_path_for(_resource_name, _resource)
+        success_messages_path
+      end
+    end
   end
 end
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -33,7 +33,7 @@ Rails.application.configure do
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
-  config.action_mailer.perform_deliveries = true
+  config.action_mailer.perform_deliveries = false
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :smtp
   config.action_mailer.perform_caching = false
--- a/config/environments/development.rb
+++ b/config/environments/development.rb
@@ -33,9 +33,9 @@ Rails.application.configure do
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
-  config.action_mailer.perform_deliveries = false
+  config.action_mailer.perform_deliveries = true
   config.action_mailer.raise_delivery_errors = true
-  config.action_mailer.delivery_method = :smtp
+  config.action_mailer.delivery_method = :letter_opener
   config.action_mailer.perform_caching = false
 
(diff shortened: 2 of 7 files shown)


config/environments/development.rb
```diff
@@ -33,7 +33,7 @@
   # Store uploaded files on the local file system (see config/storage.yml for options).
   config.active_storage.service = :local
 
-  config.action_mailer.perform_deliveries = true
+  config.action_mailer.perform_deliveries = false
   config.action_mailer.raise_delivery_errors = true
   config.action_mailer.delivery_method = :smtp
```
malparty:
```suggestion
  config.action_mailer.perform_deliveries = true
  config.action_mailer.raise_delivery_errors = true
  config.action_mailer.delivery_method = :letter_opener
```

I recommend using the gem `letter_opener` as this will allow you to previous the emails in your localhost ;-)

app/controllers/api/v1/confirmations_controller.rb
```diff
@@ -2,6 +2,10 @@
 
 module Api
   module V1
-    class ConfirmationsController < Devise::ConfirmationsController; end
+    class ConfirmationsController < Devise::ConfirmationsController
+      def after_confirmation_path_for(_resource_name, _resource)
+        success_messages_path
```
malparty, line 7: I'm not sure, but should we use the path that generate a new token? This way we can get both a success http status AND the token allowing to start using the API directly? 

(and the bonus is that then we do not need an empty "SuccessMessageController" 🙈 )

malparty (changes requested):
Some suggestions.

We might want to add a test for the redirection after user confirmation ;-)

app/controllers/api/v1/confirmations_controller.rb
phongvhd93: Usually, when I have to verify my account, I tap on the link in a confirmation email, and it lands me on a web page with a success message. Then I must log in with my email and password to continue using the application. 

I will create another story for your suggestion and work on that.

malparty: As you prefer ;-) Just an optional suggestion.

From my experience, I'm most of the time already logged in to the service when I confirm my account, so I land on the already logged in interface :harold:

malparty (approved):

app/controllers/api/v1/confirmations_controller.rb
blyscuit: @phongvhd93 I think we can do the minimal (I also like the login via confirmation) but up to your preference 🙏 .

blyscuit (approved):

byhbt (approved):

Issue #12, opened 2023-05-29
[API] As a user, I can view my Uploaded Keywords


blyscuit:
## Why

User should be able to see their uploaded keyword.

## Acceptance Criteria

- Create **authenticated** endpoint `GET` `/api/v1/keyword`
    - Input:
        - `Authentication` Header: `Bearer {Token}`
    - Output: JSON
    
    ```jsx
    [
    	{
    		"id": 1,
    		"keyword": "1234",
    		"created_at": "2023-05-22T07:38:34.420Z",
    		"updated_at": "2023-05-22T07:38:34.420Z"
    	},
    	{
    		"id": 2,
    		"keyword": "9012",
    		"created_at": "2023-05-22T08:13:47.101Z",
    		"updated_at": "2023-05-22T08:13:47.101Z"
    	}
    ]
    ```

PR #52, opened 2023-06-20
[#11][#12] [Backend][API] As a user, I can view my uploaded keywords

### LINKED ISSUE #11: [Backend] As a user, I can view my uploaded keywords
## Why

User should be able to see their uploaded keyword.

## Acceptance Criteria

- Modify endpoint from #12 to show uploaded keywords uploaded by the user calling the endpoint.

phongvhd93:
- close:
  - #11 
  - #12

## What happened 👀

Allow a logged-in user to get their uploaded keywords.
## Insight 📝

- Update the query in the `index` action of `KeywordsController` to fetch the user's keywords based on their `user_id`
- Add 2 more fields `created_date` and `updated_date` to `KeywordSerializer`
- Write test

## Proof Of Work 📹

Screenshot 2023-06-20 at 15 59 49

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -3,7 +3,8 @@
 module Api
   module V1
     class KeywordsController < ApplicationController
-      after_action :verify_policy_scoped, only: :index
+      before_action :authorize!
+
       def index
         keywords = policy_scope(Keyword)
         render json: KeywordSerializer.new(keywords).serializable_hash.to_json
@@ -22,6 +23,10 @@ module Api
       end
 
       private
+
+      def authorize!
+        authorize :keyword
+      end
 
       def csv_form
         @csv_form ||= CsvUploadForm.new(current_user)
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -4,10 +4,8 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        first_keyword = Keyword.new({ id: 1, name: 'First keyword' })
-        second_keyword = Keyword.new({ id: 2, name: 'Second keyword' })
-
-        render json: KeywordSerializer.new([first_keyword, second_keyword]).serializable_hash.to_json
+        keywords = Keyword.where(user_id: current_user.id)
+        render json: KeywordSerializer.new(keywords).serializable_hash.to_json
       end
 
       def create
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -4,7 +4,7 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        keywords = Keyword.where(user_id: current_user.id)
+        keywords = Keyword.where(user: current_user)
         render json: KeywordSerializer.new(keywords).serializable_hash.to_json
       end
 
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -3,8 +3,9 @@
 module Api
   module V1
     class KeywordsController < ApplicationController
+      after_action :verify_policy_scoped, only: :index
       def index
-        keywords = Keyword.where(user: current_user)
+        keywords = policy_scope(Keyword)
         render json: KeywordSerializer.new(keywords).serializable_hash.to_json
       end
 
--- a/app/policies/application_policy.rb
+++ b/app/policies/application_policy.rb
@@ -1,0 +1,53 @@
+# frozen_string_literal: true
+
+class ApplicationPolicy
+  attr_reader :user, :record
+
+  def initialize(user, record)
+    @user = user
+    @record = record
+  end
+
+  def index?
+    false
+  end
+
+  def show?
+    false
+  end
+
+  def create?
+    false
+  end
+
+  def new?
+    create?
+  end
+
+  def update?
+    false
+  end
+
+  def edit?
+    update?
+  end
+
+  def destroy?
+    false
+  end
+
+  class Scope
+    def initialize(user, scope)
+      @user = user
+      @scope = scope
+    end
+
+    def resolve
+      raise NotImplementedError, "You must define #resolve in #{self.class}"
+    end
+
+    private
+
+    attr_reader :user, :scope
+  end
+end
--- a/app/policies/keyword_policy.rb
+++ b/app/policies/keyword_policy.rb
@@ -6,4 +6,18 @@ class KeywordPolicy < ApplicationPolicy
       scope.where(keywords: { user_id: user.id })
     end
   end
+
+  def index?
+    user_valid?
+  end
+
+  def create?
+    user_valid?
+  end
+
+  private
+
+  def user_valid?
+    !user.id.nil? && !user.email.nil?
+  end
 end
--- a/app/policies/keyword_policy.rb
+++ b/app/policies/keyword_policy.rb
@@ -18,6 +18,6 @@ class KeywordPolicy < ApplicationPolicy
   private
 
   def user_valid?
-    !user.id.nil? && !user.email.nil?
+    user && !user.id.nil? && !user.email.nil?
   end
 end
--- a/app/policies/keyword_policy.rb
+++ b/app/policies/keyword_policy.rb
@@ -1,0 +1,9 @@
+# frozen_string_literal: true
+
+class KeywordPolicy < ApplicationPolicy
+  class Scope < Scope
+    def resolve
+      scope.where(keywords: { user_id: user.id })
+    end
+  end
+end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -63,6 +63,7 @@ ActiveRecord::Schema.define(version: 2023_06_13_043615) do
     t.datetime "confirmed_at", precision: 6
     t.datetime "confirmation_sent_at", precision: 6
     t.string "unconfirmed_email"
+    t.index ["confirmation_token"], name: "index_users_on_confirmation_token", unique: true
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
   end
--- a/db/schema.rb
+++ b/db/schema.rb
@@ -63,7 +63,6 @@ ActiveRecord::Schema.define(version: 2023_06_13_043615) do
     t.datetime "confirmed_at", precision: 6
     t.datetime "confirmation_sent_at", precision: 6
     t.string "unconfirmed_email"
-    t.index ["confirmation_token"], name: "index_users_on_confirmation_token", unique: true
     t.index ["email"], name: "index_users_on_email", unique: true
     t.index ["reset_password_token"], name: "index_users_on_reset_password_token", unique: true
   end
--- a/spec/policies/keyword_policy_spec.rb
+++ b/spec/policies/keyword_policy_spec.rb
@@ -24,6 +24,30 @@ RSpec.describe KeywordPolicy, type: :request do
           expect(scope.to_a).to be_empty
         end
       end
+
+      context 'given the user is a keyword owner' do
+        it 'permits :create, :index' do
+          user = Fabricate(:user)
+          keyword = Fabricate(:keyword, user: user)
+
+          policy = described_class.new(user, keyword)
+
+          expect(policy).to permit_action(:create)
+          expect(policy).to permit_action(:index)
+        end
+      end
+    end
+
+    context 'given the user is an anonymous user' do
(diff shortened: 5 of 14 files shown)


spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -1,20 +1,50 @@
 # frozen_string_literal: true
 
+require 'byebug'
```
byhbt: I think we don't need the `byebug` here right?

```suggestion
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -1,20 +1,50 @@
...
+
+        get api_v1_keywords_path, headers: create_token_header
+
+        expect(response).to have_http_status(:success)
+      end
 
-      keywords = JSON.parse(response.body, symbolize_names: true)[:data]
+      it 'returns an empty array' do
+        user = Fabricate(:user)
+        Fabricate.times(3, :keyword, user_id: user.id)
```
byhbt: It took a bit time to understand this 😅 , please declare things explicitly. And for this case, we also don't need to `fabricate` 3 items, it could save us some query. 

```suggestion
        another_user = Fabricate(:user)
        Fabricate.times(:keyword, user_id: another_user.id)
        
        get api_v1_keywords_path, headers: create_token_header
        
        response_body = JSON.parse(response.body, symbolize_names: true)
        expect(response_body[:data].count).to eq 0
```

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -4,10 +4,8 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        first_keyword = Keyword.new({ id: 1, name: 'First keyword' })
-        second_keyword = Keyword.new({ id: 2, name: 'Second keyword' })
-
-        render json: KeywordSerializer.new([first_keyword, second_keyword]).serializable_hash.to_json
+        keywords = Keyword.where(user_id: current_user.id)
```
byhbt: Can we 💭 ? If possible i prefer passing an object to the Active query.

```suggestion
        keywords = Keyword.where(user: current_user)
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -1,20 +1,50 @@
...
+
+        get api_v1_keywords_path, headers: create_token_header(user)
+
+        expect(response).to have_http_status(:success)
+      end
+
+      it 'returns three items' do
+        user = Fabricate(:user)
+        Fabricate.times(3, :keyword, user_id: user.id)
```
blyscuit: Are we allow to fabricate with `user` instead of `user.id` as well?

spec/requests/api/v1/keywords_controller_spec.rb
blyscuit: It should be?
```
        Fabricate(:keyword, user_id: another_user.id)
```

blyscuit (commented):
Looks good.

db/schema.rb
```diff
@@ -63,7 +63,6 @@
     t.datetime "confirmed_at", precision: 6
     t.datetime "confirmation_sent_at", precision: 6
     t.string "unconfirmed_email"
-    t.index ["confirmation_token"], name: "index_users_on_confirmation_token", unique: true
```
byhbt: hi! @phongvhd93 i am not sure why we have this change here? As we don't have any change in the migration.

spec/requests/api/v1/keywords_controller_spec.rb
byhbt: yup, my typo issue.

spec/requests/api/v1/keywords_controller_spec.rb
byhbt: yup, that's much cleaner

```suggestion
        Fabricate.times(3, :keyword, user: user)
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -1,20 +1,49 @@
 # frozen_string_literal: true
 
+require 'rails_helper'
+
 RSpec.describe Api::V1::KeywordsController, type: :request do
   describe 'GET#index' do
-    it 'returns success status' do
-      get api_v1_keywords_path, headers: create_token_header
+    context 'given a logged in user gets theirs keyword' do
+      it 'returns success status' do
```
byhbt: I think we can improve the spec here a bit more structure like below

```suggestion
    context 'given a logged in' do
      context 'given the user query their own keyword' do
        it 'returns success status' do
          # ..
        end
      end

      context 'given the user has no keyword' do
        it 'returns success status' do
          # ..
        end
      end
    end
    
    context 'given non-logged in user' do
    
    end
```

spec/requests/api/v1/keywords_controller_spec.rb
malparty: As a side note: be aware that when building an instance (`Fabricate.build`) it will only assign the attributes that are declared. If you use `user: user`, it means the `user_id` will not be assigned. 
So depending on what your code is doing, you might need one or another. 

This should also be taken into consideration when writing the Fabricator. 

In Flexi, we ended up with the rule of thumb: 
- always use `instances` (user) — to ensure 99% of the cases will be consistent and to avoid duplicate declarations
- only use `foreign key id` (user.id) when there is no choice (e.g. a form that parses attributes expect ids, not instances).

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -4,10 +4,8 @@ module Api
   module V1
     class KeywordsController < ApplicationController
       def index
-        first_keyword = Keyword.new({ id: 1, name: 'First keyword' })
-        second_keyword = Keyword.new({ id: 2, name: 'Second keyword' })
-
-        render json: KeywordSerializer.new([first_keyword, second_keyword]).serializable_hash.to_json
+        keywords = Keyword.where(user: current_user)
```
malparty: I would recommend using the [Pundit gem](https://github.com/varvet/pundit/) because it there is a lot of chance that your next Web project requires permissions/policies management. 

On the other hand, Pundit has a very nice documentation (just the README.md file) and you can find examples of usage in the previous client projects ;-)

db/schema.rb
phongvhd93: Yes, that's weird, I didn't touch that file. It seems the system updated the file itself. I will revert this change.

malparty: `schema.rb` is auto-generated by Rails. It will reflect the current state of the DB based on the migrations that were run. 

This often occurs when you start your project in BRANCH_A (that includes new migrations). Then you switch branches with the project still running, so schema.rb is overwritten by Rails and the new branch has a pending change on `shcema.rb`. If you commit your code too quickly you might include the unwanted changes of `schema.rb` in the new branch — no worries, it happens all the time to everyone :harold:

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -1,20 +1,59 @@
 # frozen_string_literal: true
 
+require 'rails_helper'
+
 RSpec.describe Api::V1::KeywordsController, type: :request do
   describe 'GET#index' do
-    it 'returns success status' do
-      get api_v1_keywords_path, headers: create_token_header
+    context 'given a logged in user' do
+      context 'given the user query query their own keyword' do
```
```suggestion
      context 'given the user has keywords' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -1,20 +1,59 @@
...
+
+          response_body = JSON.parse(response.body, symbolize_names: true)
+          expect(response_body[:data].count).to eq 3
+        end
+      end
+
+      context 'given the user has no keyword' do
+        it 'returns success status' do
+          user = Fabricate(:user)
+          Fabricate(:keyword, user: user)
```
```suggestion
          _another_user_keyword = Fabricate(:keyword)
```

(ensure your keyword fabricator Fabricates a user when not provided)

malparty (approved):
Really clean now with the Policy!! ✨ 

Last minor suggestions, approved ahead! 🚀

blyscuit (approved):

app/policies/keyword_policy.rb
byhbt, line 1: We also need to write a spec for this policy as well @phongvhd93 .

phongvhd93: @byhbt, I've written test for the policy in fccfcac

app/policies/application_policy.rb
```diff
@@ -0,0 +1,53 @@
...
+    false
+  end
+
+  def edit?
+    update?
+  end
+
+  def destroy?
+    false
+  end
```
byhbt, line 37: You can write a test for these methods:

```
  context 'given the user is a keyword owner' do
    it 'permits :new, :create, :index' do
      user = Fabricate(:user)
      keyword = Fabricate(:keyword, user: user)

      policy = described_class.new(user, keyword)

      expect(policy).to permit_action(:new)
      expect(policy).to permit_action(:create)
      expect(policy).to permit_action(:index)
      # ... and other method here
    end
  end

  context 'given the user is an anonymous user' do
    it 'does NOT permit :new, :create, :index' do
      anonymous_user = NullUser.new
      keyword = Fabricate(:keyword)

      policy = described_class.new(anonymous_user, weekly_days_off)

      expect(policy).not_to permit_action(:new)
      expect(policy).not_to permit_action(:create)
      expect(policy).not_to permit_action(:index)
      # ... and other method here
    end
  end
```

What is the `NullUser`?
https://thoughtbot.com/blog/rails-refactoring-example-introduce-null-object

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -3,11 +3,10 @@
 module Api
   module V1
     class KeywordsController < ApplicationController
+      after_action :verify_policy_scoped, only: :index
```
byhbt: Usually we use:

```
before_action :authorize!


  private

  def authorize!
    authorize :keyword
  end
```

spec/policies/keyword_policy_spec.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KeywordPolicy, type: :request do
```
byhbt, line 5: You can write a test for these methods:

```ruby
  context 'given the user is a keyword owner' do
    it 'permits :new, :create, :index' do
      user = Fabricate(:user)
      keyword = Fabricate(:keyword, user: user)

      policy = described_class.new(user, keyword)

      expect(policy).to permit_action(:new)
      expect(policy).to permit_action(:create)
      expect(policy).to permit_action(:index)
      # ... and other method here
    end
  end

  context 'given the user is an anonymous user' do
    it 'does NOT permit :new, :create, :index' do
      anonymous_user = NullUser.new
      keyword = Fabricate(:keyword)

      policy = described_class.new(anonymous_user, weekly_days_off)

      expect(policy).not_to permit_action(:new)
      expect(policy).not_to permit_action(:create)
      expect(policy).not_to permit_action(:index)
      # ... and other method here
    end
  end
```

What is the `NullUser`?
https://thoughtbot.com/blog/rails-refactoring-example-introduce-null-object

phongvhd93: fixed in b30d09e

app/controllers/api/v1/keywords_controller.rb
phongvhd93: fixed in b30d09e

spec/policies/keyword_policy_spec.rb
```diff
@@ -0,0 +1,53 @@
...
+      it 'does NOT permit :create, :index' do
+        anonymous_user = NullUser.new
+        keyword = Fabricate(:keyword)
+
+        policy = described_class.new(anonymous_user, keyword)
+
+        expect(policy).not_to permit_action(:create)
+        expect(policy).not_to permit_action(:index)
+      end
+    end
```
byhbt: i think these test should be outside the `describe 'Scope' do` block. Please check the example again. 
`describe 'Scope' do` for testing the nested class `class Scope < Scope` within the `KeywordPolicy`

phongvhd93: fixed in f4b7c35

byhbt (approved):

app/policies/keyword_policy.rb
```diff
@@ -0,0 +1,23 @@
...
+  end
+
+  def create?
+    user_valid?
+  end
+
+  private
+
+  def user_valid?
+    user && !user.id.nil? && !user.email.nil?
```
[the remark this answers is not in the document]
blyscuit: How about, from this [link](https://blog.shiftasia.com/better-null-check-with-safe-navigation-operator-in-ruby/). For explicit boolean, can keep as is.
```suggestion
    user&.id && user&.email
```

PR #53, opened 2023-06-21
[#16] [Backend] Save Searched keywords data from Google to Database

### LINKED ISSUE #16: [Backend] Save Searched keywords data from Google to Database
## Why

## Acceptance Criteria

- From Google search result, find the information for:

blyscuit:
- close #16

## What happened 👀

When user upload a CSV file, get Google Search result for the keyword and save it to database.

## Insight 📝

- Created `SearchKeywordJob` to run the search and save data for one keyword.
- Added DB class to update `Keyword`.
- Added `SearchKeywordError` for error result.
- Not sure how to solve line limits yet.

## Proof Of Work 📹

Screen Shot 2023-06-21 at 18 59 02
Screen Shot 2023-06-21 at 19 00 29

### DIFF
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -11,7 +11,7 @@ module Api
       end
 
       def create
-        if csv_form.save(params[:file])
+        if csv_form.save(params[:file], source_name(params))
           render json: create_success_response
         else
           render_errors(
@@ -36,6 +36,12 @@ module Api
         {
           meta: I18n.t('csv.upload_success')
         }
+      end
+
+      def source_name(params)
+        # TODO: Read source_name from parameter
+        
+        'Google'
       end
     end
   end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -11,7 +11,11 @@ module Api
       end
 
       def create
-        if csv_form.save(params[:file])
+        saved_keywords = csv_form.save(params[:file])
+        if saved_keywords
+          # TODO: Create background job for all keywords
+          Google::SearchKeywordJob.perform_now(saved_keywords.first)
+
           render json: create_success_response
         else
           render_errors(
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -11,11 +11,7 @@ module Api
       end
 
       def create
-        saved_keywords = csv_form.save(params[:file])
-        if saved_keywords
-          # TODO: Create background job for all keywords
-          Google::SearchKeywordJob.perform_now(saved_keywords.first)
-
+        if csv_form.save(params[:file])
           render json: create_success_response
         else
           render_errors(
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -38,9 +38,9 @@ module Api
         }
       end
 
-      def source_name(params)
+      def source_name(_params)
         # TODO: Read source_name from parameter
-        
+
         'Google'
       end
     end
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -13,21 +13,23 @@ class CsvUploadForm
     @current_user = current_user
   end
 
-  def save(file)
+  def save(file, source_name)
     @file = file
     return false unless file && valid?
 
-    saved_keywords = []
-    ActiveRecord::Base.transaction { saved_keywords += keywords.each(&:save) }
+    ActiveRecord::Base.transaction {
+      source = Source.find_or_create_by({ name: source_name })
+      keywords(source).each(&:save)
+    }
 
     errors.empty?
   end
 
   private
 
-  def keywords
+  def keywords(source)
     CSV.read(file).filter_map do |row_columns|
-      keyword = Keyword.new(name: row_columns.to_csv.chomp, user_id: @current_user.id)
+      keyword = Keyword.new(name: row_columns.to_csv.chomp, user_id: @current_user.id, source: source)
 
       keyword if keyword.valid?
     end
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -17,9 +17,10 @@ class CsvUploadForm
     @file = file
     return false unless file && valid?
 
-    ActiveRecord::Base.transaction { keywords.each(&:save) }
+    saved_keywords = []
+    ActiveRecord::Base.transaction { saved_keywords += keywords.each(&:save) }
 
-    errors.empty?
+    errors.empty? ? saved_keywords : false
   end
 
   private
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -17,10 +17,10 @@ class CsvUploadForm
     @file = file
     return false unless file && valid?
 
-    ActiveRecord::Base.transaction {
+    ActiveRecord::Base.transaction do
       source = Source.find_or_create_by({ name: source_name })
       keywords(source).each(&:save)
-    }
+    end
 
     errors.empty?
   end
--- a/app/forms/csv_upload_form.rb
+++ b/app/forms/csv_upload_form.rb
@@ -20,7 +20,7 @@ class CsvUploadForm
     saved_keywords = []
     ActiveRecord::Base.transaction { saved_keywords += keywords.each(&:save) }
 
-    errors.empty? ? saved_keywords : false
+    errors.empty?
   end
 
   private
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -5,14 +5,16 @@ module Google
     queue_as :default
 
     def perform(keyword)
-      search_result = Google::SearchService.new(keyword[:name]).call()
+      search_result = Google::SearchService.new(keyword[:name]).call
+      raise Google::Errors::SearchKeywordError unless search_result
+
       update_keyword(keyword, search_result)
     end
 
     private
 
     def update_keyword(keyword, search_result)
-      source = Source.find_or_create_by({ :name => 'Google' })
+      source = Source.find_or_create_by({ name: 'Google' })
       keyword.update(
         top_ads_count: search_result[:ads_top_count],
         total_ads_count: search_result[:ads_page_count],
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -4,8 +4,9 @@ module Google
   class SearchKeywordJob < ApplicationJob
     queue_as :default
 
-    def perform(keyword)
-      search_result = Google::SearchService.new(keyword[:name]).call
+    def perform(keyword_id)
+      keyword = Keyword.find(keyword_id)
+      search_result = Google::SearchService.new(keyword.name).call
       raise Google::Errors::SearchKeywordError unless search_result
 
       update_keyword(keyword, search_result)
@@ -15,16 +16,7 @@ module Google
 
     def update_keyword(keyword, search_result)
       source = Source.find_or_create_by({ name: 'Google' })
-      keyword.update(
-        top_ads_count: search_result[:ads_top_count],
-        total_ads_count: search_result[:ads_page_count],
-        ads_links: search_result[:ads_top_urls],
-        result_count: search_result[:result_count],
-        result_links: search_result[:result_urls],
-        total_link_count: search_result[:total_link_count],
-        html: search_result[:html],
-        source: source
-      )
+      keyword.update(search_result.merge(source: source))
     end
   end
 end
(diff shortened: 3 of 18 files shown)


app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Google
+  class SearchKeywordJob < ApplicationJob
+    queue_as :default
+
+    def perform(keyword)
+      search_result = Google::SearchService.new(keyword[:name]).call
```
byhbt: Usually when passing data to the background job, we pass the ID instead of the whole object. 

```suggestion
    def perform(keyword_id)
      keyword = Keyword.find(keyword_id)
      search_result = Google::SearchService.new(keyword.name).call
```

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,30 @@
...
+      raise Google::Errors::SearchKeywordError unless search_result
+
+      update_keyword(keyword, search_result)
```
byhbt, line 8: Let reduce the magic string by moving it as a constant of the `Google::SearchService`. or `Source::GOOGLE`
The pass it to the update method, instead of querying it there. For example, if we have another search service `Bing` then it easier to modify. 

```suggestion
      google_source = Source.find_or_create_by({ name: Google::SearchService::NAME })
      
      update_keyword(keyword, search_result, google_source)
```

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -0,0 +1,130 @@
...
+RSpec.describe Google::SearchKeywordJob, type: :job do
+  include ActiveJob::TestHelper
+
+  describe '#perform' do
+    context 'given a valid request' do
+      it 'saves 3 as top_ads_count in the Database' do
+        stub_request(:get, %r{google.com/search}).to_return(body: file_fixture('html/valid_google.html').read)
+        keyword = Fabricate(:keyword)
+
+        described_class.perform_now keyword
```
byhbt: For testing the job, you can reference this guide.  https://guides.rubyonrails.org/testing.html#testing-jobs

app/jobs/google/search_keyword_job.rb
malparty: This is more than just a convention: Background jobs are enqueues (e.g. redis will store the reference to the job and the arguments passed into it). 
To enqueue a job, the arguments are then serialized and stored.
Passing the whole object can have multiple issues:
- slow queue performance as the enqueued jobs are bigger
- serialize/deserialize might miss the associated objects
- the state of the object in DB might have changed at the time the job is running
- ...

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -11,7 +11,11 @@ def index
       end
 
       def create
-        if csv_form.save(params[:file])
+        saved_keywords = csv_form.save(params[:file])
+        if saved_keywords
+          # TODO: Create background job for all keywords
```
```suggestion
          # TODO: Perform "later" and for "all" keywords
```

If you call `perform_now`, it won't run async and will expose your request to timeout.
https://api.rubyonrails.org/classes/ActiveJob/Execution.html#method-i-perform_now

In the final version we need to use [perform_later](https://api.rubyonrails.org/classes/ActiveJob/Enqueuing/ClassMethods.html#method-i-perform_later) instead ;-)

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -11,7 +11,11 @@ def index
       end
 
       def create
```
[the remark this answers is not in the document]
malparty: I recommend to move the Job enqueuing to an ActiveRecord callback: That's a good use case to explore the [Active Record Callbacks guide](https://guides.rubyonrails.org/active_record_callbacks.html) 😎 

The idea is to let the controller simply "save" the keywords, without caring about anything else. 
The `model` will then know that each time a keyword is saved, it must "enqueue" the SearchKeyword job 😇 

This would also resolve your current todo of `enqueue "ALL" keywords) :)

app/forms/csv_upload_form.rb
```diff
@@ -17,9 +17,10 @@ def save(file)
     @file = file
     return false unless file && valid?
 
-    ActiveRecord::Base.transaction { keywords.each(&:save) }
+    saved_keywords = []
+    ActiveRecord::Base.transaction { saved_keywords += keywords.each(&:save) }
 
-    errors.empty?
+    errors.empty? ? saved_keywords : false
```
malparty: If you agree with the "Active Record callback" usggestion, we do not need this change.

Also please remember the part about returning `boolean` or `object / nil`, but avoiding to return an object OR a boolean in the same function 🥴

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,30 @@
...
+        top_ads_count: search_result[:ads_top_count],
+        total_ads_count: search_result[:ads_page_count],
+        ads_links: search_result[:ads_top_urls],
+        result_count: search_result[:result_count],
+        result_links: search_result[:result_urls],
+        total_link_count: search_result[:total_link_count],
+        html: search_result[:html],
+        source: source
+      )
+    end
```
malparty:
```suggestion
    def update_keyword(keyword, search_result)
      keyword.update search_attributes
    end
    
    def search_attributes
      {
        top_ads_count: search_result[:ads_top_count],
        total_ads_count: search_result[:ads_page_count],
        ads_links: search_result[:ads_top_urls],
        result_count: search_result[:result_count],
        result_links: search_result[:result_urls],
        total_link_count: search_result[:total_link_count],
        html: search_result[:html],
        source: Source.find_or_create_by({ name: 'Google' })
      }
    end
```

The `lines` limit is not a big problem, as long as the long-methods are a single hash/array/... i.e. they have a very specific responsibility.

In this case, I would recommend extracting the hash build into a dedicated method (even if this dedicated method would still be more than 10 lines, it would be OK).


## But I have another question: 

> Why do we have different keys in `keyword` and `search_result`?

For example, in the line: `top_ads_count: search_result[:ads_top_count],`, why `top_ads_count` is not the same as `ads_top_count`? 🤔 

If we adopt the same keys (*which I do recommend!*), we could write the code in a more elegant way: 

```
    def update_keyword(keyword, search_result)
      source = Source.find_or_create_by({ name: 'Google' })

      keyword.update(search_result.merge(source: source))
    end
```

In case `search_result` contains some unwanted keys, you can either `except` them, or `slice` all the desired keys using a constant that lists all the keys we want to use. But I think `except/slice` is not needed in that case 🤔

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,30 @@
+# frozen_string_literal: true
+
+module Google
+  class SearchKeywordJob < ApplicationJob
+    queue_as :default
+
+    def perform(keyword)
+      search_result = Google::SearchService.new(keyword[:name]).call
+      raise Google::Errors::SearchKeywordError unless search_result
```
malparty: I would not recommend raising an error in this case. Raising an error is useful when you do not know what to do about the error (e.g. when coding a library and want the user to be noticed). But in Ruby, most of the use cases implement a "return false" or "return nil". 
As we are in a Job that updates a keyword, if we can't get any search results, let's just update the keyword `status` column to `failed` and end the job without raising any error :)

malparty (changes requested):
First suggestions :) 
I'll review tests only once these suggestions are implemented as tests will change 🙇

app/jobs/google/search_keyword_job.rb
blyscuit: @byhbt I am not sure what you mean by "moving it as a constant of the Google::SearchService" but the benefit of "have another search service Bing then it easier to modify" sounds good, could you elaborate more on the example 🙏

byhbt: Yup, here is the idea:

This is `SearchKeywordJob` which calls the `Google::SearchService`. However, the limitation is in the `update_keyword`, the `source` we `find_or_create_by` is `Google`.

If one day, we don't want to search by the `Google::SearchService`, we would like to use `Bing::SearchService` instead, then we need to go to the `update_keyword`, update the string `{ name: 'Bing' }` there as well. 

So instead of that, the `update_keyword` method should care about what search are using, they just accept one more param `source` `update_keyword(keyword, search_result, google_source)`. So it could be easier to change right? 💭  We don't need to update the `Google` word multiple places.

blyscuit: I moved the `source_name` to controller level so that `Keyword.callback action` knows which source to call. So this file had changed somewhat 🙏

app/forms/csv_upload_form.rb
blyscuit: I migrated to Action Record Callback. Thank you for the suggestion. 🙏

app/jobs/google/search_keyword_job.rb
blyscuit: Updated this to allow easier call on the job.

app/jobs/google/search_keyword_job.rb
blyscuit: Added `status` and `failed` status to `Keyword`. 👍

spec/jobs/google/search_keyword_job_spec.rb
blyscuit: Thank you, I added to the controller spec. Will make sure to use it in the Background job task 👍

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,23 @@
...
+      keyword = Keyword.find(keyword_id)
+      search_result = Google::SearchService.new(keyword.name).call
+      update_keyword(keyword, search_result)
+    end
+
+    private
+
+    def update_keyword(keyword, search_result)
+      if search_result
+        keyword.update search_result.merge(status: 1)
```
byhbt: Let use enum instead, we try to avoid the magic number / magic string by reference to the model constant or enum.
Please update other place as well

```suggestion
        keyword.update search_result.merge(status: Keyword.status[:parsed)
```

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -0,0 +1,119 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Google::SearchKeywordJob, type: :job do
+  include ActiveJob::TestHelper
+
+  describe '#perform' do
+    context 'given a valid request' do
+      it 'saves 3 as ads_top_count in the Database' do
```
byhbt: I think we can reduce this phrase, because it's quite obvious. 🎨 

```suggestion
      it 'saves 3 as ads_top_count' do
```

db/migrate/20230620095911_add_search_to_keyword.rb
```diff
@@ -0,0 +1,15 @@
...
+    change_table :keywords do |t|
+      t.integer :ads_top_count, null: true
+      t.integer :ads_page_count, null: true
+      t.string :ads_top_urls, array: true, null: true
+      t.integer :result_count, null: true
+      t.string :result_urls, array: true, null: true
+      t.integer :total_link_count, null: true
+      t.string :html, null: true
+      t.references :source, null: true, foreign_key: true
+      t.integer :status, default: 0, null: false
```
byhbt: How about set the default 0 to the `integer` field? I think it's more strict than just leave it as `null`. Basically, if the given record is has `ads_top_count` is `null`, it also can be consider as the `0`

Let separate the `references` column with other column.

```suggestion
      t.integer :ads_top_count, default: 0, null: false
      t.integer :ads_page_count, default: 0, null: false
      t.string :ads_top_urls, array: true, null: false
      t.integer :result_count, default: 0, null: false
      t.string :result_urls, array: true, null: true
      t.integer :total_link_count, default: 0, null: false
      t.string :html, null: true
      t.integer :status, default: 0, null: false
      
      t.references :source, null: true, foreign_key: true
```

app/jobs/google/search_keyword_job.rb
```diff
@@ -0,0 +1,23 @@
+# frozen_string_literal: true
+
+module Google
+  class SearchKeywordJob < ApplicationJob
+    queue_as :default
```
byhbt:
```suggestion
```

Because it's default, i think we can remove this line. Explicit is good i think, however if it's a convention in Rails, then we can eliminate this to avoid typing mistake.

app/jobs/google/search_keyword_job.rb
malparty: Small additional note:

if you use string enum on the model: `enum status: { in_progress: 'in_progress', parsed: 'parsed', failed: 'failed' }` then you can have a shorter yet explicit query like:

```suggestion
        keyword.update search_result.merge(status: 'parsed')
```

In terms of performance, it is true that using a small int is better than using a string, but it remains a negligible difference in the context of this application.

app/jobs/google/search_keyword_job.rb
malparty: Clean ✨

malparty (approved):
Nice changes =)
Approved ahead with Thanh's suggestion still to handle

app/jobs/google/search_keyword_job.rb
blyscuit: Updated to string.

byhbt (approved):

phongvhd93 (approved):

PR #54, opened 2023-06-23
[#14][#13] [Backend] [API] As a user, I can view my uploaded keywords with pagination

phongvhd93:
- close #13 
- close #14  

## What happened 👀

Use the gem`pagy` to implement pagination on the endpoint `keyword`

## Insight 📝

- Add new param `pagination_params`.
- Add new meta to contain current page, page size, and total number of items
- Write test for pagination.

## Proof Of Work 📹

Screenshot 2023-06-23 at 09 30 50

PR #56, opened 2023-06-23
[#18][#19] [API] As a user, I can access my keyword details

### LINKED ISSUE #18: [Backend] As a user, I can access my keyword details
## Why

The user should be able to access their uploaded keyword’s detail.

- The detail of the keyword should include:

## Acceptance Criteria

- Modify endpoint from #19 to fetching a single keyword and map the data as a JSON result.
- Keywords should belong to the user, returns error if user does not have permission to certain keyword.

### LINKED ISSUE #19: [API] As a user, I can access my keyword details
## Why

The user should be able to access their uploaded keyword’s detail.

- The detail of the keyword should include:

## Acceptance Criteria

- Create **authenticated** endpoint `GET` `/api/v1/keyword/detail`
    - Input:
        - keyword: `String`
    - Output: JSON

```jsx
{
		"id": 1,
		"keyword": "1234",
		"created_at": "2023-05-22T07:38:34.420Z",
		"updated_at": "2023-05-22T07:38:34.420Z",
    "top_adwords_count": "Number of AdWords advertisers in the top position. Added as Keyword’s variable `top_adwords_count`.",
    "total_adwords": "Total number of AdWords advertisers on the page. Added as Keyword’s variable `total_adwords`.",
    "top_adword_url": "URLs of the AdWords advertisers in the top position. Added as Keyword’s variable `top_adword_url`.",
    "search_result_count": "Number of the non-AdWords results on the page. Added as Keyword’s variable `search_result_count`.",
    "non_adwords_urls": "URLs of the non-AdWords results on the page. Database One `Keyword` <> Many `SearchUrl`.",
    "total_links": "Total number of links (all of them) on the page. Added as Keyword’s variable `total_links`.",
    "page_cache": "HTML code of the page/cache of the page. Database One `Keyword` <> One `PageCache`.",
}
```

blyscuit:
- close #18 
- close #19 

## What happened 👀

Call `GET` to `api/v1/keywords/:id` can return the keyword detail belongs to the user.

## Insight 📝

- Add `show` to KeywordController.
- Add new serializer `KeywordDetailSerializer`.
- Modify Error response to fix some bug.
- Add `Rescuable` to support future error catching.

## Proof Of Work 📹

Screen Shot 2023-06-23 at 15 18 32

### DIFF
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -10,6 +10,16 @@ module Api
 
       before_action :doorkeeper_authorize!
 
+      def raw_resources(klass, policy_scope_klass = nil)
+        resolve_policy_scope(klass, policy_scope_klass)
+      end
+
+      def resolve_policy_scope(scope, policy_scope_klass)
+        return policy_scope(scope) unless policy_scope_klass
+
+        policy_scope_klass.new(current_user, scope).resolve
+      end
+
       private
 
       def current_user
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -4,6 +4,7 @@ module Api
   module V1
     class ApplicationController < ActionController::API
       include ErrorRenderable
+      include Rescuable
       include Pagy::Backend
       include Pundit::Authorization
 
--- a/app/controllers/api/v1/application_controller.rb
+++ b/app/controllers/api/v1/application_controller.rb
@@ -10,16 +10,6 @@ module Api
 
       before_action :doorkeeper_authorize!
 
-      def raw_resources(klass, policy_scope_klass = nil)
-        resolve_policy_scope(klass, policy_scope_klass)
-      end
-
-      def resolve_policy_scope(scope, policy_scope_klass)
-        return policy_scope(scope) unless policy_scope_klass
-
-        policy_scope_klass.new(current_user, scope).resolve
-      end
-
       private
 
       def current_user
@@ -43,6 +33,16 @@ module Api
           page: params[:page],
           items: params[:per_page]
         }
+      end
+
+      def authorized_resources(klass, policy_scope_klass = nil)
+        resolve_policy_scope(klass, policy_scope_klass)
+      end
+
+      def resolve_policy_scope(scope, policy_scope_klass)
+        return policy_scope(scope) unless policy_scope_klass
+
+        policy_scope_klass.new(current_user, scope).resolve
       end
     end
   end
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -23,7 +23,7 @@ module Api
       end
 
       def show
-        keyword = raw_resources(Keyword).find(params[:id])
+        keyword = authorized_resources(Keyword).find(params[:id])
         options = { include: [:source] }
 
         render json: KeywordDetailSerializer.new(keyword, options).serializable_hash.to_json
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -23,8 +23,16 @@ module Api
       end
 
       def show
-        keyword = Keyword.find_by(id: params[:id])
-        render json: KeywordDetailSerializer.new(keyword).serializable_hash.to_json
+        keyword = Keyword.find_by(user: current_user, id: params[:id])
+
+        if keyword
+          render json: KeywordDetailSerializer.new(keyword).serializable_hash.to_json
+        else
+          render_errors(
+            details: [I18n.t('keyword.not_found')],
+            status: :not_found
+          )
+        end
       end
 
       private
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -23,17 +23,14 @@ module Api
       end
 
       def show
-        keyword = Keyword.find_by(user: current_user, id: params[:id])
-
-        if keyword
-          options = { :include => [:source] }
-          render json: KeywordDetailSerializer.new(keyword, options).serializable_hash.to_json
-        else
-          render_errors(
-            details: [I18n.t('keyword.not_found')],
-            status: :not_found
-          ) 
-        end
+        keyword = policy_scope(Keyword).find(params[:id])
+        options = { include: [:source] }
+        render json: KeywordDetailSerializer.new(keyword, options).serializable_hash.to_json
+      rescue ActiveRecord::RecordNotFound
+        render_errors(
+          details: [I18n.t('keyword.not_found')],
+          status: :not_found
+        )
       end
 
       private
--- a/app/controllers/api/v1/keywords_controller.rb
+++ b/app/controllers/api/v1/keywords_controller.rb
@@ -27,11 +27,6 @@ module Api
         options = { include: [:source] }
 
         render json: KeywordDetailSerializer.new(keyword, options).serializable_hash.to_json
-      rescue ActiveRecord::RecordNotFound
-        render_errors(
-          details: [I18n.t('keyword.not_found')],
-          status: :not_found
-        )
       end
 
       private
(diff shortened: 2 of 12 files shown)


app/controllers/api/v1/keywords_controller.rb
```diff
@@ -22,6 +22,18 @@ def create
         end
       end
 
+      def show
+        keyword = policy_scope(Keyword).find(params[:id])
```
byhbt: Can we just use?

```suggestion
        keyword = Keyword.find(params[:id])
```

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -22,6 +22,18 @@ def create
...
+        options = { include: [:source] }
+
+        render json: KeywordDetailSerializer.new(keyword, options).serializable_hash.to_json
+      rescue ActiveRecord::RecordNotFound
+        render_errors(
+          details: [I18n.t('keyword.not_found')],
+          status: :not_found
+        )
```
byhbt: Can we handle this part globally? I mean we should not `recuse` on every `show` like this when the resource does not exist.

blyscuit: @byhbt Do you mean to declare this `render_errors(` as a private variable inside this class?

> should not recuse on every show

How can we determined if the keyword exist for the particular user + keyword? The idea is to show only keyword that the user owned. 🤔

app/controllers/api/v1/keywords_controller.rb
blyscuit: The idea is to show only keyword that the user owned, hence the pundit scope. 😄

byhbt: ah right, sorry i miss it.

app/controllers/api/v1/application_controller.rb
```diff
@@ -5,9 +5,20 @@ module V1
     class ApplicationController < ActionController::API
       include Pundit::Authorization
       include ErrorRenderable
+      include Rescuable
 
       before_action :doorkeeper_authorize!
 
+      def raw_resources(klass, policy_scope_klass = nil)
```
malparty:
```suggestion
      def authorized_resources(klass, policy_scope_klass = nil)
```

The name `raw_resources` was added after we had the `resources` method and mean that it was the resources without pagination.
But if I could travel back in time, I would probably choose another name :harold:.

As this case is simpler (no search, filter, ...) I recommend to give it an explicit name like `authorized_resources` 🙇

app/controllers/api/v1/application_controller.rb
```diff
@@ -5,9 +5,20 @@ module V1
...
+        resolve_policy_scope(klass, policy_scope_klass)
+      end
+
+      def resolve_policy_scope(scope, policy_scope_klass)
+        return policy_scope(scope) unless policy_scope_klass
+
+        policy_scope_klass.new(current_user, scope).resolve
+      end
```
malparty, line 46: It seems like this method can be made private, isn't it?

app/controllers/api/v1/application_controller.rb
malparty: Also, be careful, it might conflict with https://github.com/nimblehq/ic-rails-phong-bliss/pull/54#discussion_r1243232823

app/controllers/concerns/api/v1/rescuable.rb
```diff
@@ -0,0 +1,18 @@
...
+      extend ActiveSupport::Concern
+
+      included do
+        rescue_from ActiveRecord::RecordNotFound do
+          render_errors(
+            details: [I18n.t('api.errors.not_found')],
+            status: :not_found
+          )
+        end
+      end
```
malparty:
```suggestion
      included do
        rescue_from ActiveRecord::RecordNotFound { render_not_found_error }
      end
      
      def render_not_found_error
        render_errors(
          details: [I18n.t('api.errors.not_found')],
          status: :not_found
        )
      end
```

The included bock is here to gather the DSL specific to rails. It is recommended to keep it short so that we can easily grasp how the concern is affecting the class that will implement it.

app/policies/keyword_policy.rb
```diff
@@ -15,6 +15,10 @@ def create?
     user_valid?
   end
 
+  def show?
+    user_valid?
+  end
+
   private
 
   def user_valid?
```
malparty, line 24: `user && !user.id.nil? && !user.email.nil?` 

Not sure when I missed this one 🙈 

Can we improve it a little bit? 

`user&.id.present? && user&.email.present?`

app/serializers/keyword_detail_serializer.rb
```diff
@@ -0,0 +1,13 @@
...
+class KeywordDetailSerializer < ApplicationSerializer
+  attributes :name
+  attributes :ads_top_count
+  attributes :ads_page_count
+  attributes :ads_top_urls
+  attributes :result_count
+  attributes :result_urls
+  attributes :total_link_count
+  attributes :html
+  has_one :source
```
malparty, line 13:
```suggestion

  has_one :source
```

Optional but it might be interesting to add an empty line in this case 💭

spec/fabricators/keyword_fabricator.rb
```diff
@@ -5,3 +5,15 @@
...
+Fabricator(:keyword_parsed, from: :keyword) do
+  status :parsed
+  ads_top_count { FFaker.rand 9 }
+  ads_page_count { FFaker.rand 9 }
+  ads_top_urls(rand: 9) { FFaker::Internet.http_url }
+  result_count { FFaker.rand 9 }
+  result_urls(rand: 9) { FFaker::Internet.http_url }
+  total_link_count { FFaker.rand 9 }
+  html { FFaker::HTMLIpsum.body }
+  source { Fabricate(:source) }
```
malparty, line 19:
```suggestion

  source { Fabricate(:source) }
```

Let's try to separate the associated fields from the direct fields. This improves the readability ~

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -144,4 +144,64 @@
       end
     end
   end
+
+  describe 'GET#show' do
+    context 'when CSV file is valid' do
+      it 'returns success status' do
+        keyword = Fabricate(:keyword_parsed)
+        user = keyword.user
+        get api_v1_keyword_path(keyword.id), headers: create_token_header(user)
```
malparty, line 175:
```suggestion

        get api_v1_keyword_path(keyword.id), headers: create_token_header(user)
```

Arrange

Act

Assert

🚀

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -144,4 +144,64 @@
...
+
+    context 'when there is no keyword' do
+      it 'returns not found status' do
+        get api_v1_keyword_path(0), headers: create_token_header
+
+        expect(response).to have_http_status(:not_found)
+      end
+    end
+
+    context 'given a non-logged in user' do
```
```suggestion
    context 'given no token headers' do
```
I do not see any "unlogged user" in the test below 🙈

malparty (changes requested):
Let's also ensure about the consistency of naming between #54 and this PR 🙌

app/controllers/concerns/api/v1/rescuable.rb
blyscuit: I was not able to get `rescue_from ActiveRecord::RecordNotFound { render_not_found_error }` working with `missing end` error so I modified this to
```
        rescue_from ActiveRecord::RecordNotFound do
          render_not_found_error
        end
```

malparty: Hmm... probably due to the DSL...
Let's see what the documentation says about it... 👀 

https://apidock.com/rails/ActiveSupport/Rescuable/ClassMethods/rescue_from

oh! 
Can you try this?

```
rescue_from, :with => render_not_found_error
```

🚀

malparty:
```
rescue_from, with: render_not_found_error
```

Maybe this works too... need to try :D

malparty (approved):
Approved with a last suggestion :)

app/controllers/concerns/api/v1/rescuable.rb
blyscuit: Works great 👍

byhbt (approved):

PR #63, opened 2023-06-27
[#17] [Backend] Automatically search uploaded keywords on Google in background queue

### LINKED ISSUE #17: [Backend] Automatically search uploaded keywords on Google in background queue
## Why

## Acceptance Criteria

- Crawling should be done in the background with queue for different CSV upload from all the users.

blyscuit:
- close #17

## What happened 👀

Keyword search now perform in background.

## Insight 📝

- Add `KeywordSearchError` for when Google crawl failed.
  - Raising error for `Sidekiq` will automatically add the job to [retry queue](https://www.rubydoc.info/gems/sidekiq/Sidekiq/JobRetry) default to 25 retries.
  - Update Keyword status to `failed`.

## Proof Of Work 📹

Screen Shot 2023-06-27 at 16 36 27
Screen Shot 2023-06-27 at 16 37 39

Background demo:

https://github.com/nimblehq/ic-rails-phong-bliss/assets/6356137/90ed5127-0f74-4d67-8190-085d3e4e389e

### DIFF
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -16,7 +16,7 @@ module Google
       else
         keyword.update({ status: :failed })
 
-        raise Google::Errors::SearchKeywordError unless search_result
+        raise Google::Errors::SearchKeywordError
       end
     end
   end
--- a/app/jobs/google/search_keyword_job.rb
+++ b/app/jobs/google/search_keyword_job.rb
@@ -15,6 +15,8 @@ module Google
         keyword.update search_result.merge(status: :parsed)
       else
         keyword.update({ status: :failed })
+
+        raise Google::Errors::SearchKeywordError unless search_result
       end
     end
   end
--- a/spec/jobs/google/search_keyword_job_spec.rb
+++ b/spec/jobs/google/search_keyword_job_spec.rb
@@ -88,6 +88,7 @@ RSpec.describe Google::SearchKeywordJob, type: :job do
 
         described_class.perform_now keyword.id
         keyword.reload
+      rescue Google::Errors::SearchKeywordError
 
         expect(
           [
@@ -112,6 +113,7 @@ RSpec.describe Google::SearchKeywordJob, type: :job do
         keyword = Fabricate(:keyword)
 
         described_class.perform_now keyword.id
+      rescue Google::Errors::SearchKeywordError
 
         expect(keyword.reload.status).to eq('failed')
       end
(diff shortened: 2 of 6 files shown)


byhbt (approved):

app/jobs/google/search_keyword_job.rb
```diff
@@ -15,6 +15,8 @@ def update_keyword(keyword, search_result)
         keyword.update search_result.merge(status: :parsed)
       else
         keyword.update({ status: :failed })
+
+        raise Google::Errors::SearchKeywordError unless search_result
```
malparty:
```suggestion
        raise Google::Errors::SearchKeywordError
```

This condition seems duplicated with the `if/else` (as is is already in the `else` block that ensures `search_result` is falsy. 

But why exactly do we want to raise this error? 👀 
I do not see any reason why we need to raise an error here: as we already handle the case by setting the keyword status to `failed`, it seems to me that we do not need to consume more resources with raising an error. 🤔

blyscuit: @malparty Sidekiq comes with a [built-in retry](https://www.rubydoc.info/gems/sidekiq/Sidekiq/JobRetry) feature. By raising an error in the job, Sidekiq will create a retry queue for that job. We set status fail so that users and backend know it is already fetched but fail. 🙏

malparty: Oh yeah makes sense.

Don't forget to remove the useless `unless`  condition though! :D 

```suggestion
        raise Google::Errors::SearchKeywordError
```

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -133,9 +136,19 @@
         allow(described_class).to receive(:perform_now)
 
         described_class.perform_now keyword_id
+      rescue Google::Errors::SearchKeywordError
 
         expect(described_class).to have_received(:perform_now).with(keyword_id).exactly(:once)
       end
+
+      it 'raises SearchKeywordError' do
```
malparty:
```suggestion
      it 'raises SearchKeywordError to allow Sidekiq to retry the search' do
```

Should we document this implicit important concept? :)

malparty (approved):
Approved with a last suggestion :)

app/jobs/google/search_keyword_job.rb
blyscuit: Nice catch.

PR #65, opened 2023-06-28
[#64] [Chore] Source.name should be unique and case insensitive

### LINKED ISSUE #64: [Chore] `Source.name` should be unique and case insensitive
## Why

Source should not have duplicate name.

## Acceptance Criteria

- Add `unique` tag to `Source.name`.
- Add case insensitive tag to `Source.name`.

blyscuit:
- close #64 

## What happened 👀

Make `Source.name` unique and case insensitive.

## Insight 📝

- Create `AddUniqueToSources` DB migration.
- Add `case-sensitive: false` to `Source`.

## Proof Of Work 📹

Test passes and App behaves the same.

### DIFF
--- a/db/migrate/20230628044856_add_unique_to_sources.rb
+++ b/db/migrate/20230628044856_add_unique_to_sources.rb
@@ -1,0 +1,6 @@
+class AddUniqueToSources < ActiveRecord::Migration[7.0]
+  def change
+    change_column :sources, :name, :string, unique: true, null: false
+    add_index :sources, :name, unique: true
+  end
+end
--- a/db/migrate/20230628044856_add_unique_to_sources.rb
+++ b/db/migrate/20230628044856_add_unique_to_sources.rb
@@ -1,6 +1,0 @@
-class AddUniqueToSources < ActiveRecord::Migration[7.0]
-  def change
-    change_column :sources, :name, :string, unique: true, null: false
-    add_index :sources, :name, unique: true
-  end
-end
--- a/spec/forms/csv_upload_form_spec.rb
+++ b/spec/forms/csv_upload_form_spec.rb
@@ -72,8 +72,9 @@ RSpec.describe CsvUploadForm, type: :form do
     context 'given a file with 9 keywords and 1 blank' do
       it 'saves 9 keywords with the correct user' do
         user = Fabricate(:user)
+        source = Fabricate(:source)
         form = described_class.new(user)
-        form.save(file_fixture('csv/blank.csv'), nil)
+        form.save(file_fixture('csv/blank.csv'), source.name)
 
         expect(keyword_for_user(user).count).to eq(9)
       end
--- a/spec/models/source_spec.rb
+++ b/spec/models/source_spec.rb
@@ -4,22 +4,20 @@ require 'rails_helper'
 
 RSpec.describe Source, type: :model do
   context 'when creating two Sources with different cases' do
-    it 'does NOT create a second source' do
-      described_class.find_or_create_by({ name: 'nimble' })
-
+    it 'saves 1 source' do
       expect do
+        described_class.find_or_create_by({ name: 'nimble' })
         described_class.find_or_create_by({ name: 'nImBle' })
-      end.not_to change(described_class, :count)
+      end.to change(described_class, :count).by(1)
     end
   end
 
   context 'when creating two Sources with different names' do
     it 'saves 2 sources' do
-      described_class.find_or_create_by({ name: 'nimble1' })
-
       expect do
+        described_class.find_or_create_by({ name: 'nimble1' })
         described_class.find_or_create_by({ name: 'nimble2' })
-      end.to change(described_class, :count).by(1)
+      end.to change(described_class, :count).by(2)
     end
   end
 end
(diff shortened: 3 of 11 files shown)


byhbt (approved):

db/migrate/20230628044856_add_unique_to_sources.rb
```diff
@@ -0,0 +1,6 @@
+class AddUniqueToSources < ActiveRecord::Migration[7.0]
+  def change
+    change_column :sources, :name, :string, unique: true, null: false
+    add_index :sources, :name, unique: true
+  end
+end
```
malparty: These migrations are quite dangerous if you do not have a "data migration" to run before.
In case your DB contains duplicated names, the migration will fail to run and so the application will not start.

Here is an example of migration from our current project:

```
# frozen_string_literal: true

class AddUniqueIndexOnNameToTenants < ActiveRecord::Migration[7.0]
  def up
    update_existing_tenants

    remove_index :tenants, :name
    add_index :tenants, :name, unique: true, where: 'discarded_at IS NULL'
  end

  def down
    remove_index :tenants, :name
    add_index :tenants, :name
  end

  private

  def update_existing_tenants
    Tenant.with_discarded.group_by { |tenant| tenant.name.downcase }
                         .each do |_name, tenants|
      next unless tenants.size > 1

      tenants.each_with_index do |tenant, index|
        tenant.name = "#{tenant.name} #{index + 1}"
        tenant.save!(validate: false)
      end
    end
  end
end
```

spec/forms/csv_upload_form_spec.rb
```diff
@@ -72,8 +72,9 @@
     context 'given a file with 9 keywords and 1 blank' do
       it 'saves 9 keywords with the correct user' do
         user = Fabricate(:user)
+        source = Fabricate(:source)
```
malparty, line 75: Once the constraint is added, there is a high chance that running twice `Fabricate(:source)` or just running once `Fabricate.times(2, :source)` will raise the UNIQUE violation error 🙈 

For that, I recommend to update the Fabricator using the [FFaker](https://github.com/ffaker/ffaker) and its `.unique` util so that it would always return a unique name.

```
name { FFaker::Company.unique.name }
```

spec/models/source_spec.rb
```diff
@@ -0,0 +1,25 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe Source, type: :model do
+  context 'when saving two Sources with different cases' do
+    it 'saves only one Source' do
```
malparty:
```suggestion
    it 'does NOT create a second source' do
```

Because the test assert `not_to changes`.

Same for the next test, it assert `to change count by 1`, so we cannot describe the test with `saves 2 sources`, instead, we can use `creates a new source`

spec/models/source_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+      described_class.find_or_create_by({ name: 'nimble' })
+
+      expect do
+        described_class.find_or_create_by({ name: 'nImBle' })
+      end.not_to change(described_class, :count)
+    end
+  end
+
+  context 'when saving two Sources with different names' do
```
malparty:
```suggestion
  context 'when creating two Sources with different names' do
```

We can `save` a change (e.g. when updating the values of attributes) without changing the count 😇

malparty (changes requested):

db/migrate/20230628044856_add_unique_to_sources.rb
blyscuit: I added the `update_existing` logic to the migration in https://github.com/nimblehq/ic-rails-phong-bliss/pull/65/commits/1dc37ad49806641f8310f8229f4724d24bbfbebc. I still use change because the index `name` did not exist before this migration.

spec/models/source_spec.rb
```diff
@@ -0,0 +1,25 @@
...
+
+  context 'when creating two Sources with different names' do
+    it 'saves 2 sources' do
+      described_class.find_or_create_by({ name: 'nimble1' })
+
+      expect do
+        described_class.find_or_create_by({ name: 'nimble2' })
+      end.to change(described_class, :count).by(1)
+    end
+  end
```
malparty: As already said, the description and the tests are different.

Either we adjust the test to suit with the description:

```suggestion
  context 'when creating two Sources with different names' do
    it 'saves 2 sources' do
      expect do
        described_class.find_or_create_by({ name: 'nimble1' })
        described_class.find_or_create_by({ name: 'nimble2' })
      end.to change(described_class, :count).by(2)
    end
  end
```

Or we adjust the description to suit with the test:

```suggestion
context 'given a source "nimble1"' do
  context 'when creating a second source "nimble2"' do
    it 'creates 1 new source' do
      described_class.find_or_create_by({ name: 'nimble1' })

      expect do
        described_class.find_or_create_by({ name: 'nimble2' })
      end.to change(described_class, :count).by(1)
    end
  end
end
```


The point is just: you cannot have a test `it saves 2` that `expect count change by 1` 🙈

malparty (approved):
Approved ahead with a last need for change 🙇

spec/models/source_spec.rb
blyscuit: The first example makes a lot of senses. 👍

Issue #58, opened 2023-06-26
[Backend] Save search result from Bing to database


phongvhd93:
## Why

After crawling the data from Bing successfully, we must store the search result in the database for later use.

## Acceptance Criteria

Save the following information from Bing search results to database:
- Number of AdWords advertisers in the top position.
- Total number of AdWords advertisers on the page.
- URLs of the AdWords advertisers in the top position.
- Number of the non-AdWords results on the page.
- URLs of the non-AdWords results on the page.
- Total number of links (all of them) on the page
- HTML code of the page/cache of the page.

PR #66, opened 2023-06-28
[#58][#57] [Backend] As a user, I can upload a list of keywords to trigger a Bing search

### LINKED ISSUE #55: [Backend] As a user, I can upload a list of keywords to trigger a Bing search
## Why

The system should automatically crawl Bing search for the following data.

- For each search result/keyword result page on Bing, you will store in the database the following information on the first page of results:

## Acceptance Criteria

- Add backend function for accessing Bing search page with the Keywords
- This function should be called when user is uploading a CSV file.

### LINKED ISSUE #57: [Backend] Automatically search uploaded keywords on Bing in background queue
# Why
The system should automatically crawl Bing for the following data.

- For each search result/keyword result page on Bing, you will store in the database the following information on the first page of results:
# Acceptance Criteria
Crawling should be done in the background with queue for different CSV upload from all the users.

phongvhd93:
- close #55 
- close #57
## What happened 👀

The system should automatically crawl Bing search for the uploaded keywords and save the results to the Database.

## Insight 📝

- Move `SearchKeywordJob` out of `Google` module to reuse it in Bing search.
- Update the logic in `SearchKeywordJob` to check the search source.
- Add param `source` for the `create` action in the `keywords_controller`.
- Write tests.

## Proof Of Work 📹

Screenshot 2023-06-28 at 17 08 09


app/controllers/api/v1/keywords_controller.rb
```diff
@@ -38,10 +38,8 @@ def create_success_response
         }
       end
 
-      def source_name(_params)
-        # TODO: Read source_name from parameter
-
-        'Google'
+      def source_name(params)
```
```suggestion
      def source_param
```

- methods that return the value of a param usually end with `_param`.
- as the `params` variable is an instance variable, then we do not need to pass it as an argument ;-)

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,26 @@
+# frozen_string_literal: true
+
+class SearchKeywordJob < ApplicationJob
+  def perform(keyword_id)
+    keyword = Keyword.includes(:source).where(keywords: { id: keyword_id }).first
```
malparty:
```suggestion
    keyword = Keyword.includes(:source).find_by(keywords: { id: keyword_id })
```

And please try, maybe a simple `find` would also work in that case:

```
    keyword = Keyword.includes(:source).find(keyword_id)
```

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,26 @@
...
+    keyword_name = keyword.name
+    source_name = keyword.source.name.downcase
+    case source_name
+    when 'google'
+      search_result = Google::SearchService.new(keyword_name).call
+    when 'bing'
+      search_result = Bing::SearchService.new(keyword_name).call
+    end
```
malparty:
```suggestion
    search_result = search_service(keyword).new(keyword_name).call
```

```
private

def search_service(keyword)
  source_name = keyword.source.name.downcase

  return Bing::SearchService if source_name == 'bing'

  Google::SearchService
end
```

What about we extract the concern that retrieves the service class into a dedicated method?
As you might have seen Ruby loves "small methods" :D

spec/fabricators/keyword_fabricator.rb
```diff
@@ -4,4 +4,5 @@
   id { FFaker::Identification.ssn }
   name { FFaker::FreedomIpsum.word }
   user { Fabricate(:user) }
```
malparty, line 7:
```suggestion

  user { Fabricate(:user) }
```

spec/fabricators/source_fabricator.rb
```diff
@@ -1,5 +1,5 @@
 # frozen_string_literal: true
 
 Fabricator(:source) do
-  name 'MyWebsite'
+  name 'Google'
```
malparty:
```suggestion
  name { ['Google', 'Bing'].sample }
```

This allow test to be exposed to more cases (unless the test explicitly declares which source to use) and therefore identifies some issues that could stay hidden otherwise. 

Also, when reading the Fabricator, we get a better insight about what are the supported values :)

malparty (changes requested):

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,26 @@
...
+  end
+
+  private
+
+  def update_keyword(keyword, search_result)
+    if search_result
+      keyword.update search_result.merge(status: :parsed)
+    else
+      keyword.update({ status: :failed })
+    end
```
byhbt:
```suggestion
    keyword_data = search_result ? search_result.merge(status: :parsed) : { status: :failed }
    keyword.update(keyword_data)
```

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -2,139 +2,277 @@
...
+            ]
+          ).to all(be_nil)
+          expect(
+            [
+              keyword.ads_top_count,
+              keyword.ads_page_count,
+              keyword.result_count,
+              keyword.total_link_count
+            ]
+          ).to all(eq(0))
```
byhbt: Please extract the expectation like this, it could improve the readability 🖌️ 

```suggestion
          keyword_urls = [ keyword.ads_top_urls, keyword.result_urls, keyword.html ]
          expect(keyword_urls).to all(be_nil)
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -58,89 +58,179 @@
...
-          end
+      context 'when CSV file is missing' do
+        it 'returns unprocessable_entity status' do
+          post api_v1_keywords_path, headers: create_token_header
 
-          expect(Keyword.where(name: 'Apple').first[:ads_top_count]).to eq(0)
+          expect(response).to have_http_status(:unprocessable_entity)
+        end
+
+        it 'returns a invalid_file error' do
```
byhbt: Please keep the context consistent when using `a`, `an`, `the` in the context.

```suggestion
        it 'returns invalid_file error' do
```

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -2,139 +2,277 @@
...
+          expect(keyword.reload.status).to eq('failed')
+        end
+
+        it 'does not set the html attribute' do
+          stub_request(:get, %r{bing.com/search}).to_return(status: 422)
+          keyword = Fabricate(:keyword, source: Fabricate(:source, name: 'Bing'))
+
+          described_class.perform_now keyword.id
+
+          expect(keyword.reload.reload.html).not_to be_present
```
byhbt:
```suggestion
          expect(keyword.reload.html).not_to be_present
```

app/jobs/search_keyword_job.rb
byhbt: That's a good case for Factory design pattern

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -2,139 +2,277 @@
...
-        expect(keyword.reload.total_link_count).to eq(14)
-      end
+          expect(keyword.reload.total_link_count).to eq(14)
+        end
 
-      it 'saves returned html as html' do
-        html = file_fixture('html/valid_google.html').read
-        stub_request(:get, %r{google.com/search}).to_return(body: html)
-        keyword = Fabricate(:keyword)
+        it 'saves returned html as html' do
```
byhbt: Same for below context line 200. Let try to keep the context clear.

```suggestion
        it 'saves html response as html' do
```

spec/jobs/google/search_keyword_job_spec.rb
```diff
@@ -2,139 +2,277 @@
...
+          expect(keyword.reload.status).to eq('failed')
+        end
+
+        it 'does not set the html attribute' do
+          stub_request(:get, %r{google.com/search}).to_return(status: 422)
+          keyword = Fabricate(:keyword)
+
+          described_class.perform_now keyword.id
+
+          expect(keyword.reload.reload.html).not_to be_present
```
byhbt:
```suggestion
          expect(keyword.reload.html).not_to be_present
```

spec/requests/api/v1/keywords_controller_spec.rb
phongvhd93: This should be `returns an invalid_file error`, right?

spec/fabricators/keyword_fabricator.rb
phongvhd93: solved in c69e184

app/jobs/search_keyword_job.rb
phongvhd93: solved in c69e184

spec/requests/api/v1/keywords_controller_spec.rb
phongvhd93: solved in c69e184

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class SearchKeywordJob < ApplicationJob
+  def perform(keyword_id)
+    keyword = Keyword.includes(:source).find(keyword_id)
```
byhbt: It would be safer if we check the existence of the keyword for the background job.

```suggestion
    keyword = Keyword.includes(:source).find(keyword_id)
    return unless keyword

```

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,24 @@
...
+    keyword_name = keyword.name
+    search_result = search_service(keyword).new(keyword_name).call
+    update_keyword(keyword, search_result)
+  end
+
+  private
+
+  def search_service(keyword)
+    source_name = keyword.source.name.downcase
+    return Bing::SearchService if source_name == 'bing'
```
byhbt: In Rails, when comparing the string like this, we prefer to use symbol to save the memory. https://stackoverflow.com/questions/16621073/when-to-use-symbols-instead-of-strings-in-ruby

```suggestion
    source_name = keyword.source.name.to_sym
    return Bing::SearchService if source_name == :bing
```

https://stackoverflow.com/questions/16621073/when-to-use-symbols-instead-of-strings-in-ruby

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,24 @@
+# frozen_string_literal: true
+
+class SearchKeywordJob < ApplicationJob
+  def perform(keyword_id)
+    keyword = Keyword.includes(:source).find(keyword_id)
+    keyword_name = keyword.name
+    search_result = search_service(keyword).new(keyword_name).call
```
byhbt: Can we just call the service here without the param?

We can see the duplication when passing the `keyword` instance and the `keyword_name` right after that. 

```suggestion
    search_result = search_service.new(keyword_name).call
```

So you can put the keyword to the instance variable

```
attr_reader :keyword
```

Then in the `def search_service` we can get it to detect the correspondence search service.

byhbt (changes requested):

app/jobs/search_keyword_job.rb
malparty: @byhbt I would be careful in this case as the model does not have any validations on "name" to ensure is "safe" to convert and compare as a symbol 🤔 
Symbols are better for enum strings, constants, param/attributes names, ... as these are fully in the control of the code. The source name might be a string input from an API.

byhbt: right, this perspective also makes sense.

app/jobs/search_keyword_job.rb
malparty:
```suggestion
    search_result = search_service).call
```

Good point Thanh! That makes me think that we can also initialize the service in the `search_service` method too ^^ 

(while Thanh suggestion makes total sense, consider my answer as optional - feel free to decline my suggestion ;-) )

malparty (approved):
Approved ahead with Thanh's last suggestions 🙌

spec/fixtures/files/csv/valid.csv
```diff
@@ -8,3 +8,4 @@ Tranquility
 Enigma
 Melody
 Whimsical, oh
+VPN
```
blyscuit, line 11: Curious why do we need to add this Keyword. Would the spec have less change if this is not added?

phongvhd93: Because the `valid_bing.html` is an example with the keyword VPN, so I added it. Other keywords won't produce any ads result.

app/jobs/search_keyword_job.rb
phongvhd93: solved in ff0228d

malparty: @phongvhd93 I think you put `.call`  at the wrong place: `search_result = search_service` feels wrong, it should be `search_result = search_service.call` (i.e. moving `.call` to the parent method, removing it from the `search_service` method as the later must return an instance of the search service 😇 )

spec/fixtures/files/csv/valid.csv
blyscuit: In the network stub, if we return `valid_bing.html` always then there will be no difference which word we put here 🤔 . I think stubbing for one html result it is already enough as the code would not know if the html return is related to the keyword or not.

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,26 @@
...
+    search_result = search_service
+    update_keyword(search_result)
+  end
+
+  private
+
+  def search_service
+    source_name = keyword.source.name.downcase
+    keyword_name = keyword.name
+    return Bing::SearchService.new(keyword_name).call if source_name == 'bing'
```
byhbt:
```suggestion
    keyword_name = keyword.name

    return Bing::SearchService.new(keyword_name).call if source_name == 'bing'
```

byhbt (approved):
Approve in advance with a small format issue before the return

spec/fixtures/files/csv/valid.csv
phongvhd93: Yep, however I still want to it consistent 😁

blyscuit:
Seems like we are missing #58 in this PR 🙏

blyscuit:
Screen Shot 2023-07-10 at 12 38 46
Bing searching working great 👍👍👍

PR #69, opened 2023-07-03
[#20][#21][Backend][API] As a user, I can view how many URLs contain the word {word} in AdWords

### LINKED ISSUE #20: [Backend] As a user, I can view how many URLs contain the word {word} in AdWords
## Why

A user can filter how many URLs contain a specific word in the Adwords URLs

## Acceptance Criteria

- A user can only filter the search results he uploaded based on the authorization token
- Use the existing endpoint to filter: GET `api/v1/search_results?filter[adwords_url_contains]={word}`
  - Header: Authorization (bearer token)
  - Request: filter[adwords_url_contains]={word}
  - Response: List of search results
- Write test

### LINKED ISSUE #21: [API] As a user, I can view how many URLs contain the word {word} in AdWords
## Why

User should be able to see “How many URLs contain the {word} in AdWords”.

## Acceptance Criteria

- Create **authenticated** endpoint `GET` `api/v1/search_results?filter[adwords_url_contains]={word}`
    - Input:
        - word: `String`
    - Output: JSON

the `ads_top_urls` must contain the matched URLs only

```json
{
    "data": [
        {
            "id": "349",
            "type": "keyword",
            "attributes": {
                "name": "VPN",
                "created_at": "2023-07-02T14:34:00.958Z",
                "updated_at": "2023-07-02T14:34:04.072Z",
                "ads_top_urls": [
                    "https://www.expressvpn.com/go/home",
                    "https://www.expressvpn.com/go/vpn-software/vpn-windows-1",
                    "https://www.expressvpn.com/go/vpn-software/vpn-mac-1",
                    "https://www.expressvpn.com/go/vpn-server",
                    "https://vpnpros.com/compare_generic_all_d.html"
                ]
            }
        },
        {
            "id": "356",
            "type": "keyword",
            "attributes": {
                "name": "VPN",
                "created_at": "2023-07-03T03:10:04.071Z",
                "updated_at": "2023-07-03T03:10:06.021Z",
                "ads_top_urls": [
                    "https://www.top10vpn.com/best-vpn-for-vietnam/",
                    "https://www.top10vpn.com/top10/free-vpn-for-windows/v/sitelink/?bsid=c36se1kw222",
                    "https://www.top10vpn.com/top10/free-vpn-for-mac/v/sitelink/?bsid=c36se1kw222",
                    "https://www.top10vpn.com/top10/free-trials/v/sitelink/?bsid=c36se1kw222",
                    "https://www.expressvpn.com/go/home",
                    "https://www.vpn-mentors.com/best-free-vpn-trials/"
                ]
            }
        }
    ],
    "meta": {
        "page": 1,
        "per_page": 10,
        "total_items": 2
    }
}
```

phongvhd93:
- close #20
- close #21 

## What happened 👀

Updated the endpoint `api/v1/keywords?filter[adwords_url_contains]={word}` for querying a specific word in `ad_world_url`

## Insight 📝

- Create `KeywordQuery` and `KeywordPresenter` to handle filtering logic.
- Update the logic in `KeywordsController` to handle filtering matching adwords url.
- Write tests.

## Proof Of Work 📹

Screenshot 2023-07-14 at 10 42 21


app/controllers/api/v1/search_results_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SearchResultsController < ApplicationController
+      before_action :authorize!
+
+      def index
+        word = search_params[:adwords_url_contains]
```
byhbt: Optional idea: `word` is concise but a bit generic, we can add one more word to better describing it `word_params` or `word_query`... or you can propose your own.

```suggestion
        word_params = search_params[:adwords_url_contains]
```

spec/requests/api/v1/search_results_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+            Fabricate(:keyword, ads_top_urls: ads_top_urls, user: user)
+
+            params = { filter: { adwords_url_contains: 'VPN' }, page: 1, per_page: 1 }
+            get api_v1_search_results_path, params: params, headers: create_token_header(user)
+
+            response_body = JSON.parse(response.body, symbolize_names: true)
+            expect(response_body[:data][0][:attributes][:ads_top_urls]).to eq(['https://www.thetopvpn.com', 'https://www.nordvpn.com'])
+          end
+
+          it 'retuns metadata with page = 1, per_page = 1 and total_item = 1' do
```
byhbt:
```suggestion
          it 'returns metadata with page = 1, per_page = 1 and total_item = 1' do
```

spec/requests/api/v1/search_results_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+      end
+
+      context 'given the user does not have any keyword' do
+        it 'returns an empty array' do
+          Fabricate(:keyword)
+
+          get api_v1_keywords_path, headers: create_token_header
+
+          response_body = JSON.parse(response.body, symbolize_names: true)
+          expect(response_body[:data].count).to eq 0
```
byhbt: We also can 

```suggestion
          expect(response_body[:data]).to be_empty
```

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -0,0 +1,40 @@
...
+
+      def authorize!
+        authorize :keyword
+      end
+
+      def search_params
+        params.require(:filter).permit(:adwords_url_contains)
+      end
+
+      def get_keywords_has_ads_top_urls_contains_word(word)
```
byhbt: In rails controller, the verb can be omitted in the method name

```suggestion
      def keywords_has_ads_top_urls_contains_word(word)
```

spec/requests/api/v1/search_results_controller_spec.rb
```diff
@@ -0,0 +1,98 @@
...
+          Fabricate(:keyword, ads_top_urls: ads_top_urls, user: user)
+
+          params = { filter: { adwords_url_contains: 'VPN' }, page: 1, per_page: 1 }
+          get api_v1_search_results_path, params: params, headers: create_token_header(user)
+
+          expect(response).to have_http_status(:success)
+        end
+
+        context 'given there are adword urls contain the keyword VPN' do
+          it 'retuns 2 adword urls match the word' do
```
byhbt: Does it actually return the 2 adword urls? 💭 

```suggestion
          it 'returns 2 keywords contains the matched urls' do
```

app/controllers/api/v1/search_results_controller.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SearchResultsController < ApplicationController
```
byhbt: 🎨 Should we name it like this? Because the search-results, feel like it is a search-result listing. While here we are searching the keywords which has URL contains a word (search term).

```suggestion
    class SearchKeywordsController < ApplicationController
```

app/controllers/api/v1/search_results_controller.rb
malparty, line 1: Let's use a query object in order to isolate the query concern from the controller.
That will not only make code more maintainable, but testing the code will be easier too 🚀 

https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#queries

malparty (changes requested):

app/controllers/api/v1/search_results_controller.rb
phongvhd93: added in a9fa777

app/controllers/api/v1/search_results_controller.rb
phongvhd93: solved in a9fa777

app/forms/search_keywords_form.rb
```diff
@@ -0,0 +1,47 @@
...
+  def initialize(keywords, search_params, pagination_params)
+    @keywords = keywords
+    @search_params = search_params
+    @pagination_params = pagination_params
+  end
+  # rubocop:enable Lint/MissingSuper
+
+  def search_keywords
+    pagination_params[:page] ||= 1
+    pagination_params[:items] ||= 10
```
byhbt, line 16: To handle the pagination params, we can reference this one https://github.com/nimblehq/ewa-payroll-web/blob/develop/app/controllers/concerns/paginable.rb

byhbt: Just for your information, we can handle the pagination params like this.  https://github.com/nimblehq/ewa-payroll-web/blob/develop/app/controllers/concerns/paginable.rb

app/forms/search_keywords_form.rb
```diff
@@ -0,0 +1,47 @@
+# frozen_string_literal: true
+
+class SearchKeywordsForm < ApplicationForm
+  attr_reader :keywords, :search_params, :pagination_params
+
+  # rubocop:disable Lint/MissingSuper
```
byhbt: Do we really need this disabled? 🤔

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,31 @@
+# frozen_string_literal: true
+
+class SearchKeywordJob < ApplicationJob
+  attr_reader :keyword
+
+  def perform(keyword_id)
+    @keyword = Keyword.includes(:source).find(keyword_id)
+    search_result = search_service
```
byhbt:
```suggestion
    search_result = call_search_service
```

Or you can propose a better name. Because let think `search_result = search_service` - it is a bit confusing. 🎨

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    update_keyword(search_result)
+  end
+
+  private
+
+  def search_service
+    source_name = keyword.source.name.downcase
+    keyword_name = keyword.name
```
byhbt: it's good idea to remove the duplication of the `keyword.name`. However, either way, it just call one time because on the of condition of the source_name.

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,22 @@
...
+class KeywordsQuery
+  attr_reader :keywords, :filter_params
+
+  def initialize(keywords, filter_params)
+    @keywords = keywords
+    @filter_params = filter_params
+  end
+
+  def call
+    return nil unless filter_params[:adwords_url_contains]
```
byhbt: We can simplify this.

```suggestion
    return unless filter_params[:adwords_url_contains]
```

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,22 @@
...
+
+    get_keywords_has_ads_top_urls_contains_word(filter_params[:adwords_url_contains])
+  end
+
+  private
+
+  def get_keywords_has_ads_top_urls_contains_word(word_params)
```
byhbt: is it a `word` (single word) or `word_param` (a single word comes from a params) or `word_params` (hash/array of word params)? 

```suggestion
  def get_keywords_has_ads_top_urls_contains_word(word)
```

spec/forms/search_keywords_form_spec.rb
```diff
@@ -0,0 +1,73 @@
...
+        user = Fabricate(:user)
+        filter_params = { adwords_url_contains: 'vpn' }
+        pagination_params = { page: 1, items: 2 }
+        _pagination, keywords = described_class.new(user.keywords, filter_params, pagination_params).search_keywords
+
+        expect(keywords).to be_empty
+      end
+    end
+
+    context 'when user has keyword' do
```
byhbt:
```suggestion
    context 'given a user has keywords' do
```

Or you need to update the above context at line `7` to `when...` I am fine with both but just keep it consistent.

I see you write the correct one at `spec/requests/api/v1/search_keywords_controller_spec.rb:7`

spec/forms/search_keywords_form_spec.rb
```diff
@@ -0,0 +1,73 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SearchKeywordsForm, type: :form do
+  describe '#seach_keywords' do
+    context 'given user does not have keyword' do
```
byhbt:
```suggestion
    context 'given a user does not have any keyword' do
```

spec/jobs/search_keyword_job_spec.rb
```diff
@@ -0,0 +1,281 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe SearchKeywordJob, type: :job do
+  include ActiveJob::TestHelper
+
+  describe '#perform' do
+    context 'given search engine is Google' do
```
byhbt:
```suggestion
    context 'given Google search engine' do
```

```suggestion
    context 'given search engine name is Google' do
```

app/forms/search_keywords_form.rb
```diff
@@ -0,0 +1,47 @@
...
+  def keywords_has_ads_top_urls_contains_word
+    pagination, paginated_keywords = pagy(keywords_query.call, pagination_params)
+    word_params = search_params[:adwords_url_contains]
+    filtered_paginated_keywords = paginated_keywords.map { |item| filter_unmatched_adword_urls(item, word_params) }
+    [pagination, filtered_paginated_keywords]
+  end
+
+  def filter_unmatched_adword_urls(keyword, word_params)
+    unfiltered_urls = keyword.ads_top_urls
+    keyword.ads_top_urls = unfiltered_urls.select { |item| item.downcase.include?(word_params.downcase) }
```
byhbt: We can directly modify the value `ads_top_urls` by using the `select!` method.

```suggestion
    keyword.ads_top_urls.select! { |item| item.downcase.include?(word_params.downcase) }
```

app/jobs/search_keyword_job.rb
```diff
@@ -0,0 +1,31 @@
...
+    Google::SearchService.new(keyword_name).call
+  end
+
+  def update_keyword(search_result)
+    if search_result
+      keyword.update search_result.merge(status: :parsed)
+    else
+      keyword.update({ status: :failed })
+      raise Errors::SearchKeywordError
+    end
```
byhbt:
```suggestion
    return keyword.update search_result.merge(status: :parsed) if search_result

    keyword.update({ status: :failed })
    raise Errors::SearchKeywordError
```

app/controllers/api/v1/search_keywords_controller.rb
```diff
@@ -0,0 +1,29 @@
...
+      def index
+        pagination, keywords = search_keywords_form.search_keywords
+
+        render json: KeywordSerializer.new(keywords, meta: meta_from_pagination(pagination))
+      end
+
+      private
+
+      def search_keywords_form
+        @search_keywords_form ||= SearchKeywordsForm.new(current_user.keywords, search_params, pagination_params)
```
byhbt: Instead of eager load the keyword here, we can query the keyword inside the `KeywordsQuery`.

This is about the responsibility of the class. For example:

- Controller: works like a forwarder, forward the params into the Form.
- FormObject: handle/validate the params, validate it.
- QueryObject: base on the validated params of the Form, build the query.

So that's why we should not get a list of keywords of the current user here. 

```suggestion
        @search_keywords_form ||= SearchKeywordsForm.new(current_user, search_params, pagination_params)
```

app/forms/search_keywords_form.rb
phongvhd93: The lint is unrelated, and it is annoying so I decided to disable it.

app/forms/search_keywords_form.rb
malparty, line 1: A query object might be better for this, as the purpose of form object better fit with the `create` or `update` actions.

Read more about Query Object in: https://nimblehq.co/compass/development/code-conventions/ruby/ruby-on-rails/#queries

You will find some in Flexi too, for this use case though no need for `Parms` object (used when the Query object gets too complex).

app/controllers/api/v1/search_keywords_controller.rb
malparty, line 1: Is there a reason why we do not use the `keywords#index` controller-action? 👀 

As the filtering is abstracting behind a Form/Query object, it should not be an issue to use the same endpoint :)

app/forms/search_keywords_form.rb
malparty: Why not add `super` at the end of your constructor? 🤔 
By default this will pass the arguments, if you do not have any argument you can add `super()`. 

The use case is if later `ApplicationForm` changes and must be initialized, we must not skip its constructor. In this context, it is not a big deal, as we would just "fix the code", but if `ApplicationForm` was declared in a GEM, and a new version must initialize something (with no arguments to avoid breaking changes), then we might get an error.

byhbt: yup, i just curious because in our projects we use it a lot without with disabling.

byhbt (approved):

app/forms/search_keywords_form.rb
```diff
@@ -0,0 +1,44 @@
...
+
+    keywords_has_ads_top_urls_contains_word
+  end
+
+  private
+
+  def keywords_has_ads_top_urls_contains_word
+    pagination, paginated_keywords = pagy(keywords_query.call, pagination_params)
+    word_params = search_params[:adwords_url_contains]
+    filtered_paginated_keywords = paginated_keywords.map { |item| filter_unmatched_adword_urls(item, word_params) }
```
malparty: I would recommend moving this into a presenter object:
- the presenter can have the `word_params` in its instance variable
- the presenter can expose a new method `matching_adword_urls` that will handle a logic similar to `filter_unmatched_adword_urls` without mutating the `keyword.ads_top_urls`. In general, we should not edit the value or a model field, especially if the resulting value should have a different name (e.g. `matching_adword_urls`).

As a quick draft (non-tested), it would look like this:

```
class KeywordPresenter
  # All delegations needed here...
  delegate :name, ..., to: :keyword

  def initialize(keyword)
    @keyword = keyword
  end

  def matching_adword_urls(word_param)
    keyword.ads_top_urls.select { |item| item.downcase.include?(word_params.downcase) }
  end

  private

  attr_reader :keyword
end
```


> Note that using a presenter implies not using a form :harold: If we need this logic in both the `index` (query object and presenter) and a form (like `edit`), we would need to move that method into a decorator object.

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,22 @@
...
+
+  def initialize(current_user, filter_params)
+    @current_user = current_user
+    @filter_params = filter_params
+  end
+
+  def call
+    return unless filter_params[:adwords_url_contains]
+
+    get_keywords_has_ads_top_urls_contains_word(filter_params[:adwords_url_contains])
```
```suggestion
    filter_ads_top_urls
```

You can leverage OOP by letting that method get the appropriate filter, so calling this method is easier (i.e. the responsibility of filtering the `ads_top_urls` is totally handled by the method).

app/queries/keywords_query.rb
```diff
@@ -0,0 +1,22 @@
+# frozen_string_literal: true
+
+class KeywordsQuery
+  attr_reader :current_user, :filter_params
+
+  def initialize(current_user, filter_params)
```
```suggestion
  def initialize(keywords = Keyword.all, filter_params = {})
```

A query object should work with any `scope`. We often have some previous filtering done before calling the query object:
- Authorization (pundit policies) will apply security rules
- Controller might have specific contexts (e.g. "CommunityKeywordsController" will NOT filter by `current_user` while `MyKeywordsController` would surely do)
- ...

The query object must only focus on applying filters (often defined by the request as params).


In our case, you would call the query object from the controller:

```
def index
  keywords = KeywordQuery.new(authorized_resources(Keyword), search_params)

  paginated_keywords = pagy(keywords, pagination_params)

  #...
end
```

app/controllers/api/v1/search_keywords_controller.rb
```diff
@@ -0,0 +1,29 @@
+# frozen_string_literal: true
+
+module Api
+  module V1
+    class SearchKeywordsController < ApplicationController
+      before_action :authorize!
+
+      def index
+        pagination, keywords = search_keywords_form.search_keywords
```
malparty: Filtering by `current_user` *must* be handled by Pundit policy, not by the form or the query object 🙈

malparty (changes requested):
Many comments 🙇 

At a high level, let's remember when to use what pattern:

- QueryObject: when listing a collection (`index`)
- FormObject: to present `new/edit` actions AND to handle the resource changes with the `create`/`update` actions. FormObjects are not recommended for filtering/deleting/...
- PresenterObject: Only for display only actions (`index` & `show`). Use `form object` when in need to present data specific to a form
- DecoratorObject: when a method in charge of presentation is used by another object (a mailer, a formObject, etc...)


And let's not forget to ALWAYS apply the Pundit Policy 💯

app/forms/search_keywords_form.rb
malparty: @byhbt I think we do not have the `AppilcationForm` parent class in our project ;-)

app/controllers/api/v1/search_keywords_controller.rb
phongvhd93: This was mentioned in the AC. We also have some other stories to work with search feature so I think we can group those search logic into a separate controller.

malparty: 🙈 I missed that from the AC 🥲 

As controllers should not contain logic but only basic object calls, any advanced `search logic` will never be more complex than the `keywords#index` with a single filter.

Said differently, if we have the `search_results#index` action, there is no point to keep the `keywords#index` action (because `search_results#index` without filters must behave exactly the same as `keywords#index`). 

When working in Rails, as much as possible we try to stick to the framework conventions (though I agree sometimes it is not possible). In this case, I think using `keywords#index` is better for conventions (model, views, and routes will be all consistent and that will avoid manual configuration). 

So while it's open for discussion, my recommendation sticks to `keywords#index` ^^ 

As we are in an IC, I'm also OK that we move forward with a non-conventional route, that can allow you to explore and discover what is implicit versus what is not implicit 😇 
As long as we keep in mind that this is to avoid in a client project and we keep an eye on what issues are encountered (API project might have less issues as you won't have shared views :) )

blyscuit: Good suggestion with using Keywords Controller 🙏 @phongvhd93 feel free to update AC to match your implementation 🙇 .

phongvhd93: I updated the KeywordController to follow the convention

app/controllers/api/v1/application_controller.rb
```diff
@@ -10,8 +10,14 @@ class ApplicationController < ActionController::API
 
       before_action :doorkeeper_authorize!
 
+      rescue_from ActionController::ParameterMissing, with: :handle_parameter_missing
+
       private
 
+      def handle_parameter_missing(exception)
+        render json: { error: exception.message }, status: :bad_request
```
blyscuit: You can use `render_errors`
```suggestion
render_errors(
          details: [exception.message],
          status: :bad_request
        )
```
or add this rescue in `Rescuable.rb` as well as `handle_parameter_missing` to group the rescues we have.
```
      included do
        rescue_from ActiveRecord::RecordNotFound, with: :render_not_found_error
      rescue_from ActionController::ParameterMissing, with: :handle_parameter_missing
      end
```

app/forms/application_form.rb
```diff
@@ -3,4 +3,5 @@
 class ApplicationForm
   include ActiveModel::Model
   include ActiveModel::Attributes
+  include Pagy::Backend
```
blyscuit: Do we still need this?

app/jobs/search_keyword_job.rb
```diff
@@ -5,27 +5,24 @@ class SearchKeywordJob < ApplicationJob
...
 
   private
 
-  def search_service
+  def call_search_service
     source_name = keyword.source.name.downcase
-    keyword_name = keyword.name
 
-    return Bing::SearchService.new(keyword_name).call if source_name == 'bing'
+    return Bing::SearchService.new(keyword.name).call if source_name == 'bing'
```
blyscuit, line 17: Originally this would not have the lint error `Calls 'keyword.name' 2 times` 🤔 are you facing a different issue here?

config/application.rb
```diff
@@ -6,7 +6,7 @@
 # you've limited to :test, :development, or :production.
 Bundler.require(*Rails.groups)
 
-module GoogleSearchRuby
+module GoogleSearchRub
```
blyscuit: 🤔

app/jobs/search_keyword_job.rb
phongvhd93: Following the logic, it is only called once. So we can skip it. Thanh is already commented about this.

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -31,6 +34,20 @@ def show
 
       private
 
+      def matching_adswords_urls
+        keywords_query = KeywordsQuery.new(Keyword, filter_params)
+        pagination, keywords = paginated_authorized(keywords_query.call)
+        keyword_presenters = keywords.map { |item| KeywordPresenter.new(item, filter_params) }
+
+        render json: KeywordSerializer.new(keyword_presenters, meta: meta_from_pagination(pagination))
```
blyscuit, line 38: What do you think for queried Keywords we return a different Serializer than `all_keywords`. This way when we filter we have a different result than all?

malparty: I was about to suggest the opposite 🙈 
What if we use the KeywordsQuery object also when there are no filters? So we do not need to have branched conditions in the controller (controllers should have the least logic possible).

app/controllers/api/v1/keywords_controller.rb
```diff
@@ -31,6 +34,20 @@ def show
...
+      end
+
+      def all_keywords
+        pagination, keywords = paginated_authorized(Keyword)
+
+        render json: KeywordSerializer.new(keywords, meta: meta_from_pagination(pagination))
+      end
```
```suggestion
      def filtered_keywords
        keywords_query = KeywordsQuery.new(Keyword, filter_params)
        pagination, keywords = paginated_authorized(keywords_query.call)
        keyword_presenters = keywords.map { |item| KeywordPresenter.new(item, filter_params) }

        render json: KeywordSerializer.new(keyword_presenters, meta: meta_from_pagination(pagination))
      end
```

- rename the method to describe what it does without assuming what the "filters" are (later we might add more filters, it's the responsibility of the query object to care about what filter can be used).
- remove the `all_keyworsr` method: the query object must return all keywords if there is no filter provided.

app/controllers/api/v1/keywords_controller.rb
phongvhd93: I did think about a different Serializer for all keywords and the queried ones. However, if a client wants some extra information like the name, id, or a full array of ads_top_urls (unfiltered), creating a new serializer will require the client to handle complex mapping.  If we return an extra `matching_adword_urls` client only need to create a new property in the model and mark it as optional.

spec/jobs/search_keyword_job_spec.rb
```diff
@@ -6,7 +6,7 @@
   include ActiveJob::TestHelper
 
   describe '#perform' do
-    context 'given search engine is Google' do
+    context 'given Google is search engine' do
```
malparty: This will always be true... at least as long as Google remains a search engine.

Should we keep the previous wording? Or maybe improve it now that search engines have been renamed `source`? :)

```suggestion
    context 'given the source is Google' do
```

spec/presenters/keyword_presenter_spec.rb
```diff
@@ -0,0 +1,40 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KeywordPresenter do
+  describe '#keyword' do
```
malparty:
```suggestion
  describe '#matching_adword_urls' do
```

It seems that we are not testing the `keyword` attribute, but instead, the method `matching_adword_urls`.

spec/presenters/keyword_presenter_spec.rb
```diff
@@ -0,0 +1,40 @@
...
+        keyword = Fabricate(:keyword, ads_top_urls: ads_top_urls)
+        filter_params = { adwords_url_contains: 'vpn' }
+        result = described_class.new(keyword, filter_params).matching_adword_urls
+
+        expect(result).to eq(['https://www.thetopvpn.com', 'https://www.nordvpn.com'])
+      end
+    end
+
+    context 'given the urls do not contain the keyword vpn' do
+      it 'returns matching_adword_urls as empty array' do
```
malparty:
```suggestion
      it 'is empty' do
```

If you agree with the renaming of `describe`, we can make this one much easier 🚀

spec/queries/keywords_query_spec.rb
```diff
@@ -0,0 +1,33 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KeywordsQuery, type: :query do
+  describe '#call' do
+    context 'given user has keyword' do
+      context 'given ads_top_urls contain a word vpn' do
```
malparty:
```suggestion
      context 'given ads_top_urls contain the filtering word vpn' do
```

spec/queries/keywords_query_spec.rb
```diff
@@ -0,0 +1,33 @@
...
+          ads_top_urls = ['https://www.thetopvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+          Fabricate.times(2, :keyword, ads_top_urls: ads_top_urls, user: user)
+          filter_params = { adwords_url_contains: 'vpn' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords.count).to eq 2
+        end
+      end
+
+      context 'given ads_top_urls does not contain a word apple' do
```
malparty:
```suggestion
      context 'given ads_top_urls does not contain the filtering word apple' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -7,62 +7,156 @@
...
 
-          get api_v1_keywords_path, headers: create_token_header(user)
+            get api_v1_keywords_path, headers: create_token_header(user)
 
-          expect(response).to have_http_status(:success)
-        end
+            expect(response).to have_http_status(:success)
+          end
+
+          it 'returns three items' do
```
```suggestion
          it 'returns 3 items' do
```

As in the tests above, we use digits instead of full letters numbers 😇

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -7,62 +7,156 @@
...
+            expect(response).to have_http_status(:success)
+          end
+
+          it 'returns an empty array' do
+            Fabricate(:keyword)
+
+            get api_v1_keywords_path, headers: create_token_header
+
+            response_body = JSON.parse(response.body, symbolize_names: true)
+            expect(response_body[:data].count).to eq 0
```
```suggestion
            expect(response_body[:data]).to eq []
```

or 

```suggestion
            expect(response_body[:data]).to be_empty
```

In that second case, we must change the description to "it is empty" because an empty result can be any type of collection, not only an array :)

malparty (changes requested):

malparty (approved):

PR #70, opened 2023-07-03
[#48][#49] [API] As a user, I can Register with a Gmail Account

### LINKED ISSUE #48: [Backend] As a user, I can Register with a Gmail Account
## Why

User can register with a Gmail Account to ease the process of registering instead of using an email.

## Acceptance Criteria

- Create backend service for connecting with Google OAuth:
    1. Read `auth_code` from input.
    2. Call [`https://www.googleapis.com/oauth2/v3/token`](https://www.googleapis.com/oauth2/v3/token) with `client_id` and `client_secret` to convert `auth_code` to `acesss_token`.
    3. Use `access_token` to get user information via [`https://www.googleapis.com/oauth2/v3/userinfo?access_token=`](https://www.googleapis.com/oauth2/v3/userinfo?access_token=).
- Create backend service to save user information to User DB with `Google` login flag.

### LINKED ISSUE #49: [API] As a user, I can Register with a Gmail Account
## Why

User can register with a Gmail Account to ease the process of registering instead of using an email.

## Acceptance Criteria

- Create endpoint `/api/v1/register_google`
- Input:
    - code: `String`: auth_code from [`[https://accounts.google.com/o/oauth2/auth?client_id=[client_id]&redirect_uri=http://localhost:3000/admins/auth/google_oauth2/callback&scope=email&response_type=code](https://accounts.google.com/o/oauth2/auth?client_id=123165145085-r55q45mrr7avr7ij4sghqt419nnrsp6v.apps.googleusercontent.com&redirect_uri=http://localhost:3000/admins/auth/google_oauth2/callback&scope=email&response_type=code)`](https://accounts.google.com/o/oauth2/auth?client_id=123165145085-r55q45mrr7avr7ij4sghqt419nnrsp6v.apps.googleusercontent.com&redirect_uri=http://localhost:3000/admins/auth/google_oauth2/callback&scope=email&response_type=code)

blyscuit:
- close #48
- close #49 

## What happened 👀

User can register with a `nimblehq` Google Account.

## Insight 📝

- Added gems:
  - `omniauth-google-oauth2`
  - `omniauth-rails_csrf_protection`
- Added `OmniauthCallbacksController` for receiving Google Login information, maps to new/old users.
- Added `OAuthFake` for faking Google OAuth data.
- Add Google Apps to Secret.
- Modify Devise to allow OAuth.

## Proof Of Work 📹

https://github.com/nimblehq/ic-rails-phong-bliss/assets/6356137/1913b56b-20f0-4eae-93e7-e3384f417fdd


Gemfile
```diff
@@ -51,6 +53,7 @@ group :development, :test do
   gem 'listen' # Listens to file modifications
   gem 'letter_opener' # Preview mail in the browser instead of sending
   gem 'ffaker' # A library for generating fake data such as names, addresses, and phone numbers
+  gem 'faker' # A library for generating more fake data
```
malparty: Should we remove FFaker then? I think it is better to only keep one, thus reducing the number of external dependencies (less maintenance) and avoiding the confusion of "which faker should I use for my use case")

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+module Api
+  module V1
+    module Users
+      class OmniauthCallbacksController < Devise::OmniauthCallbacksController
+        include ErrorRenderable
+
+        rescue_from ArgumentError, with: :render_bad_request
+
+        def google_oauth2
+          @user = User.from_omniauth(auth)
```
malparty:
```suggestion
          @user = User.from_omniauth(request.env['omniauth.auth'])
```

It seems that we only need the "auth" call once. So maybe we do not need to assign it as an instance variable and to cache it? :)

app/models/user.rb
```diff
@@ -5,13 +5,39 @@ class User < ApplicationRecord
...
+    when :email
+      return
+    end
+    find_existing_user(auth) || create_new_user(auth)
+  end
+
   has_many :keywords, inverse_of: :user, dependent: :destroy
+  enum provider: { email: 'email', google_oauth2: 'google_oauth2' }
+
+  private_class_method def self.find_existing_user(auth)
```
malparty:
```suggestion
  private_class_method def self.find_by_auth(auth)
```

As ActiveRecord already exposes the "find" method that does "find existing users" (on the User class), we might want a method name that is a little bit more specific to "auth". What about "find_by_auth"?

app/models/user.rb
```diff
@@ -5,13 +5,39 @@ class User < ApplicationRecord
...
+    where(provider: auth.provider, uid: auth.uid, email: auth.info.email).first
+  end
+
+  private_class_method def self.create_new_user(auth)
```
malparty: Same suggestion as above, what about "create_from_auth"? :)

config/locales/en.yml
```diff
@@ -31,6 +31,8 @@
 
 en:
   hello: "Hello world"
+  authentication:
+    generic_error: 'Something went wrong. Please try again later.'
```
```suggestion
  authentication:
    generic_error: 'Something went wrong. Please try again later.'
  hello: "Hello world"
```

config/routes.rb
```diff
@@ -16,6 +17,7 @@
       resources :private_items, only: :index
       resources :tokens, only: [:create]
       resources :registrations, only: [:create]
+      resources :google_users, only: [:create]
```
malparty: Let's keep an alphabetical ordering ✨

spec/forms/registration_form_spec.rb
```diff
@@ -42,6 +42,25 @@
...
+    context 'given user has entered a duplicated email with an account with google provider' do
+      it 'returns false' do
+        user = Fabricate(:user, provider: Faker::Omniauth.google[:provider])
+        params = Fabricate.attributes_for(:user, email: user.email).merge!(client_id: Fabricate(:application).uid)
+        form = described_class.new
+
+        expect(form.save(params)).to be false
+      end
+
+      it 'has "already taken" error messages' do
```
malparty:
```suggestion
      it 'has the "already taken" error message' do
```

spec/models/user_spec.rb
```diff
@@ -3,9 +3,43 @@
...
-    expect(user.email).not_to be_nil
+        expect(described_class.from_omniauth(auth)).to eq(user)
+      end
+    end
+
+    context 'given a new user' do
+      it 'creates a new user with the provided attributes' do
+        auth = OmniAuth::AuthHash.new(Faker::Omniauth.google)
+
+        user = described_class.from_omniauth(auth)
```
malparty, line 24:
```suggestion
        expect do
          described_class.from_omniauth(auth)
        end.to change(described_class, :count).by(1)
```

Based on the other test. So in case this test is failing, it can suggest the next test has a false positive :)

spec/requests/api/v1/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -0,0 +1,87 @@
...
+        )
+
+        post '/api/v1/users/auth/google_oauth2/callback'
+
+        expect(response).to have_http_status(:success)
+      end
+    end
+
+    context 'given the email is not available' do
+      it 'returns an unprocessable entity response with errors' do
```
malparty:
```suggestion
      it 'returns unprocessable_entity error' do
```

Just to be consistent with the other tests ;-)

malparty (changes requested):

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          if @user.persisted?
+            render_success
+          else
+            render_errors(
+              details: @user.errors.full_messages,
+              status: :unprocessable_entity
+            )
+          end
```
byhbt: 🎨 Just to reduce the nested code level

```suggestion
          return render_success if @user.persisted?
           
          render_errors(
            details: @user.errors.full_messages,
            status: :unprocessable_entity
          )
```

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -0,0 +1,52 @@
...
+
+          render json: { success: true }
+        end
+
+        def render_bad_request
+          render_errors(
+            details: [I18n.t('api.errors.bad_request')],
+            status: :bad_request
+          )
+        end
```
byhbt: Can we move it into the `ErrorRenderable`, because i see this is quite generic render error method, which could be used in other controller

db/migrate/20230628062200_add_provider_to_users.rb
```diff
@@ -0,0 +1,8 @@
+class AddProviderToUsers < ActiveRecord::Migration[7.0]
+  def change
+    change_table :users do |t|
+      t.string :provider, default: "email"
```
byhbt, line 4:
```suggestion
      t.string :provider, default: "email"
```

We also can define it as `jsonb` data field` with default is an array `[email]`. Later we can put more field into it if needed. However in the IC, `email` is enough.

Gemfile
blyscuit: I remove `Faker` because Phong might be using `FFaker`. Added constant OAuthFake.Google instead to solve issue with redundancy.

app/models/user.rb
blyscuit: Rubocop auto correct does not like `find_by_auth` I used `find_with_auth` instead.

malparty: The reason is to avoid codding custom methods that handle exactly the same result as `find_by(auth: auth)`. In our case it is different. Ok with the renaming `with` :harold:

spec/rails_helper.rb
```diff
@@ -10,6 +10,7 @@
 require 'webmock/rspec'
 
 Dir[Rails.root.join('spec', 'support', '**', '*.rb')].each { |f| require f }
+Dir[Rails.root.join('spec', 'lib', '**', '*.rb')].each { |f| require f }
```
malparty: Instead of adding another lib folder (which can conflict later if you must write tests for a lib). What about using a `fixture` folder inside the existing `support` folder?

image

malparty (approved):
Approved with a last comment ~

spec/rails_helper.rb
blyscuit: Great idea.

byhbt (approved):

PR #71, opened 2023-07-04
[#50] [API] As a user, I can Login with a Gmail Account

### LINKED ISSUE #50: [API] As a user, I can Login with a Gmail Account
## Why

User can login with a Gmail Account after register with a Gmail Account.

## Acceptance Criteria

- Reuse endpoint `/api/v1/tokens`
- Input:

```jsx
{
	"grant_type": "assertion",
  "provider": "google",
  "assertion": "{{auth_code}}",
	"client_id": "{{ _.client_id }}",
	"client_secret": "{{ _.client_secret }}"
}
```

- Use Google OAuth service to get user information then map it to `User.authenticate()`.
- Output:

```jsx
{
	"access_token": "",
	"token_type": "Bearer",
	"expires_in": 7200,
	"refresh_token": "",
	"created_at": 1684995376
}
```

blyscuit:
- close #50

## What happened 👀

Registered user with Google Account registration can login with a Google Account and receive a user token. 

## Insight 📝

- Return token from `Doorkeeper::Application` in OAuth callback.
- Move `redirect_url` to `ENV`.
- Add `allow_forgery_protection` to skip CSRF check for APIs register/login.

## Proof Of Work 📹

Screen Shot 2023-07-12 at 16 21 05


spec/requests/api/v1/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -17,12 +17,48 @@
           }
         )
 
-        post '/api/v1/users/auth/google_oauth2/callback'
+        post "#{api_v1_user_google_oauth2_omniauth_callback_path}?state=%7B%22client_id%22%3A%22#{application.uid}k%22%7D"
```
blyscuit, line 20: Needs to add test for email vs auth login.

spec/requests/api/v1/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -17,9 +17,46 @@
...
+        )
+
+        post api_v1_user_google_oauth2_omniauth_callback_path
+
+        expect(response).to have_http_status(:bad_request)
+        expect(JSON.parse(response.body)['errors']['details']).to include(I18n.t('api.errors.bad_request'))
+      end
+    end
+
+    context 'when Doorkeeper:Application is missing' do
```
byhbt:
```suggestion
    context 'when Doorkeeper::Application is missing' do
```

app/controllers/api/v1/users/omniauth_callbacks_controller.rb
```diff
@@ -28,10 +30,24 @@ def failure
...
+        def client_application
+          state = JSON.parse(request.env['action_dispatch.request.parameters']['state'])
+          app_client_id = state['client_id']
+
+          Doorkeeper::Application.find_by(uid: app_client_id)
+        end
+
         def render_success
-          # TODO: Render Doorkeeper Token
+          user = @user
```
malparty: This looks a little bit od :)

What about using a private attribute reader instead?

```
private

attr_reader :user
```

Then you can call `user` in any instance method of this controller.

```suggestion
```

... and we do not need this line anymore :)

spec/requests/api/v1/users/omniauth_callbacks_controller_spec.rb
```diff
@@ -17,9 +17,46 @@
           }
         )
 
-        post '/api/v1/users/auth/google_oauth2/callback'
+        post "#{api_v1_user_google_oauth2_omniauth_callback_path}?state=%7B%22client_id%22%3A%22#{application.uid}k%22%7D"
 
         expect(response).to have_http_status(:success)
+        expect(JSON.parse(response.body)['access_token']).not_to be_nil
```
malparty:
```suggestion
        expect(JSON.parse(response.body)['access_token']).not_to be_blank
```

Or `.to be_present`. As in case the test return `{ access_token: '' }`, you would have a false positive (i.e. code broken, test successful)

spec/requests/api/v1/users/omniauth_callbacks_controller_spec.rb
malparty:
```suggestion
    context 'when the state[:client_id] param is NOT a valid Doorkeeper::Application' do
```

Nice catch Thanh, I did not see this missing `:` 🔍 💪 

That said, can we have a description that is more connected with the test setup? In the test, we do not have `Doorkeeper::Application = nil`, but we have `state={client_id: :missing}`, so I recommend a test description that is matching the test setup 😇

malparty (changes requested):

byhbt (approved):

malparty (approved):

blyscuit:
@byhbt @malparty Thank you for the review on this PR. Please help review the latest update on this commit with `Production` working correctly. https://github.com/nimblehq/ic-rails-phong-bliss/pull/71/commits/5360b103e3fd7d7b789ed13dcc0b98f9b06e38ca 🙏

byhbt (approved):

malparty (approved):
All good! :)

PR #74, opened 2023-07-06
[#60] [Backend] As a user, I can download a whole search result page in a PDF file

### LINKED ISSUE #60: [Backend] As a user, I can download a whole search result page in a PDF file
## Why

Export a rendered web page from the HTML code of the search result stored in the Database to a PDF file so that a user can download and view the search result.

## Acceptance Criteria

- Create a helper that can render HTML as a webpage
- Generate a PDF that contains the content of the rendered webpage
- Write test

## Resources
Gems can be used:
- [wickedpdf](https://github.com/mileszs/wicked_pdf)
- [Grover](https://github.com/Studiosity/grover)

### LINKED ISSUE #61: [API] As a user, I can download the search result page for a specific keyword
## Why

Allow the user to download the search result page for the specific keyword as a PDF file.

## Acceptance Criteria

- Create an endpoint for downloading pdf with a specific keyword: `api/v1/downloads?keyword_id={keyword_id}`
- Only a logged-in user can access that endpoint.
- Users can only download PDF files with their own keywords.
- The endpoint will return a PDF file with HTTP status 200.
- The PDF will be generated from this [story](https://github.com/nimblehq/ic-rails-phong-bliss/issues/60)

phongvhd93:
- close #60 
- close #61 

## What happened 👀

Add new endpoints to allow user download the search result page HTML as PDF format.

## Insight 📝

- Add `Grover` and `Puppeteer` to the application.
- Create a new `downloads_controller` with the endpoint `api/v1/downloads?keyword_id={number}` 
- Write test

## Proof Of Work 📹

Screenshot 2023-07-11 at 11 31 44


Makefile
```diff
@@ -7,6 +7,7 @@ dev:
 	make env/setup
 	bundle exec ./bin/dev
 
+
```
malparty:
```suggestion
```

Let's keep one breaking line only :)

config/routes.rb
```diff
@@ -16,6 +16,7 @@
       resources :private_items, only: :index
       resources :tokens, only: [:create]
       resources :registrations, only: [:create]
+      resources :downloads, only: [:index]
```
malparty: Ordering alphabetically 😇

package.json
```diff
@@ -0,0 +1,7 @@
+{
+  "name": "ic_rails_phong_bliss",
+  "private": true,
```
malparty: I recommend adding the "engine" key that contains the compatible `node` versions, but can also contain `npm` or `yarn` versions :)

package.json
malparty, line 1: I don't see any `yarn.lock` (if using yarn) nor `package.lock` (if using npm). Did you set it up locally? 👀 

To avoid using different minor versions between workspaces, it is highly recommended to add these files in the code versioning too ;-)

malparty (changes requested):

package.json
phongvhd93: I'm configuring this project to run on heroku, I'll add the files soon

malparty (approved):

blyscuit (approved):

Gemfile
```diff
@@ -29,6 +29,8 @@ gem 'doorkeeper' # OAuth 2 provider functionality to Ruby on Rails
 gem 'rails-i18n' # Translations for Rails
 # gem 'devise-i18n' # Translations for Devise
 
+gem 'grover'
```
byhbt: Follow other existing gem definition, please add the short description after the name. 

```suggestion
gem 'grover'  # A Ruby gem to transform HTML into PDFs, PNGs or JPEGs
```

config/routes.rb
```diff
@@ -12,6 +12,7 @@
         controllers tokens: 'tokens'
         skip_controllers :authorizations, :applications, :authorized_applications, :token_info, :tokens, :confirmations
       end
+      resources :downloads, only: [:index]
```
byhbt, line 16: Just for information, not request to change, normally we can define `download` as the additional route of a specific resource.

```ruby
  resources :keywords do
    member do
      get 'download'
    end
  end
```

spec/requests/api/v1/downloads_controller_spec.rb
```diff
@@ -0,0 +1,54 @@
...
+          response_headers = response.headers
+          expect(response_headers['content-type']).to eq('application/pdf')
+          expect(response_headers['content-disposition']).to eq("attachment; filename=\"vpn.pdf\"; filename*=UTF-8''vpn.pdf")
+        end
+      end
+
+      context 'given a user does not have keywords' do
+        it 'returns not found' do
+          user = Fabricate(:user)
+          html = '<h1>Hello There!</h1>'
```
byhbt:
```suggestion
```

why do we need this `html` variable? If this is required, then we can put it in the `Fabricate(:keyword)`. Same for the line 31 above. 

I would like to suggest to remove some unused variables, so we have less line of code.

config/routes.rb
phongvhd93: Thanks for your info

spec/requests/api/v1/downloads_controller_spec.rb
phongvhd93: solved in e93caf2

byhbt (approved):

PR #78, opened 2023-07-17
[#22][#23] [Backend][API] As a user, I can view How many times a specific URL shows up in stored search results

### LINKED ISSUE #22: [Backend] As a user, I can view How many times a specific URL shows up in stored search results
## Why

User should be able to see “How many times a specific URL shows up in stored search results”.

## Acceptance Criteria

- Add backend function for filtering the urls contain the search

### LINKED ISSUE #23: [API] As a user, I can view How many times a specific URL shows up in stored search results
## Why

User should be able to see “How many times a specific URL shows up in stored search results”.

## Acceptance Criteria

- Create **authenticated** endpoint `GET` `/api/v1/keyword/filter/count_url_search`
    - Input:
        - url: `String`
    - Output: JSON

```jsx
[
	{
		"id": 1,
		"keyword": "1234",
		"created_at": "2023-05-22T07:38:34.420Z",
		"updated_at": "2023-05-22T07:38:34.420Z"
	},
	{
		"id": 2,
		"keyword": "9012",
		"created_at": "2023-05-22T08:13:47.101Z",
		"updated_at": "2023-05-22T08:13:47.101Z"
	}
]
```

blyscuit:
- close #22
- close #23

## What happened 👀

Add query for `result_urls` on top of existing `adwords_url`.

## Insight 📝

- Create `keyword_params` to manage filter related parameters.
- Add `result_url` to `KeywordQuery`.
- Add `result_url` to Keyword serializer.

## Proof Of Work 📹

Screen Shot 2023-07-17 at 12 28 42


app/models/filter.rb
```diff
@@ -0,0 +1,28 @@
+# frozen_string_literal: true
+
+class Filter
```
byhbt: I think we can use Params object to handle this, instead of the PORO Filter.

app/models/filter.rb
```diff
@@ -0,0 +1,28 @@
...
+
+  def filter_type
+    if adwords_url_contains || result_url_contains
+      NORMAL_FILTER
+    elsif word && match_at_least
+      MATCH_AT_LEAST_FILTER
+    else
+      NO_FILTER
+    end
+  end
```
malparty: Just to ensure I understand well: this means we cannot run a query with both the `adwords_url_contains` and the `words` param? 👀 

If this is the case, should we review the implementation so that it checks if a filter is present, and if yes, it would apply the filter (so we can apply both filters at the same time).

app/models/filter.rb
malparty: Both are PORO, but yes, a Params object might be clearer. You can find some examples in the flex repo.

spec/queries/keywords_query_spec.rb
```diff
@@ -4,36 +4,177 @@
 
 RSpec.describe KeywordsQuery, type: :query do
   describe '#call' do
-    context 'given the user search for ads_top_urls' do
-      context 'given ads_top_urls contain the filtering word vpn' do
-        it 'returns 2 keywords' do
+    context 'when querying with adwords_url_contains' do
```
phongvhd93, line 7: should we make to wording consistent with this

'given querying with adwords_url_contains`

spec/queries/keywords_query_spec.rb
```diff
@@ -4,36 +4,177 @@
...
+      context 'given user does not have keyword' do
```
phongvhd93: Should we make it consistent for the wording by using solely `when` or `given`?

spec/queries/keywords_query_spec.rb
```diff
@@ -4,36 +4,177 @@
...
+            keywords = described_class.new(user.keywords, filter_params).call
+
+            expect(keywords.count).to eq 2
+          end
+        end
+
+        context 'when there is no urls containing the word vpn' do
+          it 'returns empty array for keywords' do
+            user = Fabricate(:user)
+            Fabricate(:keyword, user: user)
```
phongvhd93: Actually, the user play no role in the query object, querying with user's id condition is added to the query by Pundit in the `KeywordController`, I also removed `user: user` in my PR

app/models/filter.rb
blyscuit: Updated this PR to use `keyword_params` instead of `Filter`.

phongvhd93 (approved):

app/queries/keywords_query.rb
```diff
@@ -1,43 +1,41 @@
...
-    scope.where("array_to_string(ads_top_urls, '||') ILIKE ?", "%#{word}%")
-  end
+    @keywords = scope
 
-  def filter_result_urls
-    word = filter_params[:word]
-    number_of_matches = filter_params[:match_at_least]
+    @keywords = filter_ads_top_urls if filter[:adwords_url_contains].present?
+    @keywords = filter_result_urls if filter[:result_url_contains].present?
+    @keywords = filter_match_at_least if filter[:word].present? && filter[:match_at_least].present?
```
```suggestion
    scope = filter_ads_top_urls if filter[:adwords_url_contains].present?
    scope = filter_result_urls if filter[:result_url_contains].present?
    scope = filter_match_at_least if filter[:word].present? && filter[:match_at_least].present?
```

Is there a reason we don't use `scope` everywhere? 👀

malparty (changes requested):

app/models/filter.rb
blyscuit: @malparty Is this comment still relevant? Currently there are 3 filters: `adwords_url_contains`, `results_url_contains`, and `match_at_least`. The first two can be used together while `match_at_least` can only be used alone. These are defined in `keywords_params` 🙏 .

app/queries/keywords_query.rb
blyscuit: Good catch, I first worry to change `scope` so I created `keywords`. Will revert this.

blyscuit: Fixed in https://github.com/nimblehq/ic-rails-phong-bliss/pull/78/commits/a68bd8109b511dc26a8f3ac670e0c417ed051ebc

app/params/application_params.rb
```diff
@@ -0,0 +1,18 @@
...
+    @params = params
+  end
+
+  protected
+
+  attr_reader :params
+
+  def handle_param(param_name)
+    param_value = params&.dig(param_name)
+    return if param_value.blank?
```
byhbt:
```suggestion
    param_value = params&.dig(param_name)
    
    return if param_value.blank?
```

app/params/keyword_params.rb
```diff
@@ -0,0 +1,51 @@
...
+    @word = handle_param(:word)
+    @match_at_least = handle_param(:match_at_least)
+
+    if adwords_url_contains || result_url_contains
+      url_filter
+    elsif word && match_at_least
+      match_at_least_filter
+    else
+      no_filter
+    end
```
byhbt: We prefer to use the return early instead

```suggestion
    return url_filter if adwords_url_contains || result_url_contains
    return match_at_least_filter if word && match_at_least

    no_filter
```

spec/params/keyword_params_spec.rb
```diff
@@ -0,0 +1,123 @@
+# frozen_string_literal: true
+
+require 'rails_helper'
+
+RSpec.describe KeywordParams do
+  describe '#initialize' do
+    context('when init with adwords_url_contains') do
+      it('has filter with only adwords_url_contains') do
```
byhbt: No need `(` 😸 to keep the code consistent. Please adjust other places as well.

```suggestion
    context 'when init with adwords_url_contains' do
      it 'has filter with only adwords_url_contains' do
```

spec/queries/keywords_query_spec.rb
```diff
@@ -26,19 +26,132 @@
...
+          keywords = described_class.new(Keyword, {}).call
+
+          expect(keywords).to eq Keyword
+        end
+      end
+
+      context 'when there are urls containing the word vpn' do
+        it 'returns 2 keywords' do
+          ads_top_urls = ['https://www.thetopvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+          Fabricate.times(2, :keyword, ads_top_urls: ads_top_urls)
```
byhbt, line 51: When testing something like this, let try to add another "unrelated" keyword to make sure it returns only 2
```suggestion
          ads_top_urls = ['https://www.thetopvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
          Fabricate.times(2, :keyword, ads_top_urls: ads_top_urls)     
          Fabricate(:keyword, ads_top_urls: ['https://www.vpe.com'])
```

spec/queries/keywords_query_spec.rb
```diff
@@ -26,19 +26,132 @@
...
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to be_empty
+        end
+      end
+    end
+
+    context 'given the user search by word and match at least' do
+      context 'given the user search for a word vpn and match at least 1' do
+        it 'return a keyword' do
```
byhbt:
```suggestion
        it 'returns a keyword' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -325,6 +340,64 @@
           end
         end
       end
+
+      context 'when searching for a specific keyword in result urls' do
+        context 'given there are 2 keywords with result urls contain the keyword vpn' do
+          it 'returns 2 keywords contains the matched urls' do
```
byhbt:
```suggestion
          it 'returns 2 keywords contain the matched urls' do
```

spec/queries/keywords_query_spec.rb
byhbt: And in the contexts order, this happy case should come before 2 above contexts

spec/queries/keywords_query_spec.rb
```diff
@@ -26,19 +26,132 @@
...
+
+          filter_params = { adwords_url_contains: 'vpn', result_url_contains: 'game' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to be_empty
+        end
+      end
+
+      context 'when there is no urls containing the query words' do
+        it 'returns empty array for keywords' do
```
byhbt: basically it return an empty array, then just write an empty array. Let think this will become a sentence

`when there is no urls containing the query words it 'returns an empty array'`

```suggestion
        it 'returns an empty array' do
```

blyscuit: Fixed in https://github.com/nimblehq/ic-rails-phong-bliss/pull/78/commits/e91fb7286fde0e64ca3f24e6374149c1212aeebd

spec/requests/api/v1/keywords_controller_spec.rb
blyscuit: Fixed in https://github.com/nimblehq/ic-rails-phong-bliss/pull/78/commits/e91fb7286fde0e64ca3f24e6374149c1212aeebd

spec/queries/keywords_query_spec.rb
blyscuit: Fixed in https://github.com/nimblehq/ic-rails-phong-bliss/pull/78/commits/e91fb7286fde0e64ca3f24e6374149c1212aeebd

spec/queries/keywords_query_spec.rb
blyscuit: Moved the above 2 contexts out of this block as they were not related.

spec/params/keyword_params_spec.rb
blyscuit: Fixed in https://github.com/nimblehq/ic-rails-phong-bliss/pull/78/commits/e91fb7286fde0e64ca3f24e6374149c1212aeebd

byhbt (approved):

app/models/filter.rb
malparty: All resolved with the new implementation :)

malparty (approved):
🚀

PR #79, opened 2023-07-17
[#24][#25] As a user, I can view How many keywords have URLs in stored search result with specific word and number of matches

### LINKED ISSUE #24: [Backend]  As a user, I can view How many keywords have URLs in stored search result with specific word and number of matches
## Why

Users should be able to search for result_urls containing a specific word and with a number of matches.

## Acceptance Criteria

- Add backend function for filtering the urls contain the search.

### LINKED ISSUE #25: [API]  As a user, I can view How many keywords have URLs in stored search result with specific word and number of matches
## Why

Users should be able to search for result_urls containing a specific word and with a number of matches.

## Acceptance Criteria

- Use the existing endpoint `GET` `/api/v1/keywords?filter[word]={word}&filter[match_at_least]={number}`.
- Response format: 

```json
{
    "data": [
        {
            "id": "123",
            "type": "keyword",
            "attributes": {
                "name": "Apple",
                "created_at": "2023-06-28T04:00:54.581Z",
                "updated_at": "2023-06-28T04:00:56.116Z",
                "ads_top_urls": [],
                "result_urls": [
                    "https://www.apple.com/",
                    "https://www.apple.com/vn/",
                    "https://appleid.apple.com/vn/",
                    "https://appleid.apple.com/#!&page=signin",
                    "https://vi.wikipedia.org/wiki/Apple_Inc.",
                    "https://fptshop.com.vn/apple",
                    "https://www.apple.com/uk/",
                    "https://www.apple.com.cn/",
                    "https://www.apple.com/de/",
                    "https://support.apple.com/downloads/itunes"
                ],
                "matching_result_urls": [
                    "https://appleid.apple.com/vn/",
                    "https://appleid.apple.com/#!&page=signin"
                ]
            }
        }
    ],
    "meta": {
        "page": 1,
        "per_page": 50,
        "total_items": 3
    }
}
```

phongvhd93:
- close #24 
- close #25 

## What happened 👀

Add new functionality to the existing endpoint `GET api/v1/keywords` to allow a user filters keywords has `result_urls` containing a specific word and a number of matches. 

## Insight 📝

- Add a new function to filter the `result_urls` with  a specific word and a number of matches in `KeywordsQuery`
- Add new return parameter `matching_result_urls` in keyword_serializer`
- Write test.

## Proof Of Work 📹

Screenshot 2023-07-17 at 16 42 04

### DIFF
--- a/app/presenters/keyword_presenter.rb
+++ b/app/presenters/keyword_presenter.rb
@@ -9,19 +9,23 @@ class KeywordPresenter
   end
 
   def matching_adword_urls
-    return unless defined?(filter_params[:adwords_url_contains]) && filter_params[:adwords_url_contains]
+    return unless filter_params[:adwords_url_contains]
 
-    keyword.ads_top_urls.select { |item| item.downcase.include?(filter_params[:adwords_url_contains].downcase) }
+    adwords_url_contains = filter_params[:adwords_url_contains].downcase
+    keyword.ads_top_urls.select do |item|
+      item.downcase.include?(adwords_url_contains)
+    end
   rescue NoMethodError
     nil
   end
 
   def matching_result_urls
-    return unless filter_params[:word] && filter_params[:match_at_least]
+    return if filter_params[:word].blank? || filter_params[:match_at_least].blank?
 
-    keyword.result_urls.select { |item|
-      item.downcase.match("\\w*(#{filter_params[:word]}.*){#{filter_params[:match_at_least]},}")
-    }
+    word = filter_params[:word].downcase
+    keyword.result_urls.select do |item|
+      item.downcase.match("\\w*(#{word}.*){#{filter_params[:match_at_least]},}")
+    end
   rescue NoMethodError
     nil
   end
--- a/app/presenters/keyword_presenter.rb
+++ b/app/presenters/keyword_presenter.rb
@@ -1,7 +1,7 @@
 # frozen_string_literal: true
 
 class KeywordPresenter
-  delegate :id, :name, :ads_top_urls, :created_at, :updated_at, to: :keyword
+  delegate :id, :name, :ads_top_urls, :result_urls, :created_at, :updated_at, to: :keyword
 
   def initialize(keyword, filter_params)
     @keyword = keyword
@@ -12,6 +12,18 @@ class KeywordPresenter
     return unless defined?(filter_params[:adwords_url_contains]) && filter_params[:adwords_url_contains]
 
     keyword.ads_top_urls.select { |item| item.downcase.include?(filter_params[:adwords_url_contains].downcase) }
+  rescue NoMethodError
+    nil
+  end
+
+  def matching_result_urls
+    return unless filter_params[:word] && filter_params[:match_at_least]
+
+    keyword.result_urls.select { |item|
+      item.downcase.match("\\w*(#{filter_params[:word]}.*){#{filter_params[:match_at_least]},}")
+    }
+  rescue NoMethodError
+    nil
   end
 
   private
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -9,24 +9,31 @@ class KeywordsQuery
   end
 
   def call
-    if filter_params[:adwords_url_contains]
-      filter_ads_top_urls(filter_params[:adwords_url_contains])
-    elsif filter_params[:word] && filter_params[:match_at_least]
-      filter_result_urls(filter_params[:word], filter_params[:match_at_least])
-    else
-      scope
-    end
+    return filter_ads_top_urls if filter_params[:adwords_url_contains]
+
+    return filter_result_urls if filter_params[:word] && filter_params[:match_at_least]
+
+    scope
   rescue NoMethodError
     scope
   end
 
   private
 
-  def filter_ads_top_urls(word)
+  def filter_ads_top_urls
+    word = filter_params[:adwords_url_contains]&.downcase
+
+    return scope unless word
+
     scope.where("array_to_string(ads_top_urls, '||') ILIKE ?", "%#{word}%")
   end
 
-  def filter_result_urls(word, number_of_matches)
+  def filter_result_urls
+    word = filter_params[:word]
+    number_of_matches = filter_params[:match_at_least]
+
+    return scope if word.blank? || number_of_matches.blank?
+
     sub_query = Keyword.select('sub_table.id, sub_table.user_id')
                        .from(Keyword.select('id, user_id, unnest(result_urls) as result_url'), 'sub_table')
                        .where('result_url ~* ?', "(#{word}.*){#{number_of_matches},}")
--- a/app/queries/keywords_query.rb
+++ b/app/queries/keywords_query.rb
@@ -9,14 +9,28 @@ class KeywordsQuery
   end
 
   def call
-    return scope unless defined?(filter_params[:adwords_url_contains]) && filter_params[:adwords_url_contains]
-
-    filter_ads_top_urls(filter_params[:adwords_url_contains])
+    if filter_params[:adwords_url_contains]
+      filter_ads_top_urls(filter_params[:adwords_url_contains])
+    elsif filter_params[:word] && filter_params[:match_at_least]
+      filter_result_urls(filter_params[:word], filter_params[:match_at_least])
+    else
+      scope
+    end
+  rescue NoMethodError
+    scope
   end
 
   private
 
   def filter_ads_top_urls(word)
     scope.where("array_to_string(ads_top_urls, '||') ILIKE ?", "%#{word}%")
+  end
+
+  def filter_result_urls(word, number_of_matches)
+    sub_query = Keyword.select('sub_table.id, sub_table.user_id')
+                       .from(Keyword.select('id, user_id, unnest(result_urls) as result_url'), 'sub_table')
+                       .where('result_url ~* ?', "(#{word}.*){#{number_of_matches},}")
+
+    Keyword.joins("INNER JOIN (#{sub_query.to_sql}) sub_table ON keywords.id = sub_table.id").group(:id)
   end
 end
--- a/app/serializers/keyword_serializer.rb
+++ b/app/serializers/keyword_serializer.rb
@@ -8,10 +8,10 @@ class KeywordSerializer < ApplicationSerializer
              :result_urls
 
   attribute :matching_adword_urls, if: proc { |record|
-    defined?(record.matching_adword_urls) && record.matching_adword_urls.present?
+    record&.matching_adword_urls.present?
   }
 
   attribute :matching_result_urls, if: proc { |record|
-    defined?(record.matching_result_urls) && record.matching_result_urls.present?
+    record&.matching_result_urls.present?
   }
 end
--- a/app/serializers/keyword_serializer.rb
+++ b/app/serializers/keyword_serializer.rb
@@ -4,9 +4,14 @@ class KeywordSerializer < ApplicationSerializer
   attributes :name,
              :created_at,
              :updated_at,
-             :ads_top_urls
+             :ads_top_urls,
+             :result_urls
 
   attribute :matching_adword_urls, if: proc { |record|
     defined?(record.matching_adword_urls) && record.matching_adword_urls.present?
+  }
+
+  attribute :matching_result_urls, if: proc { |record|
+    defined?(record.matching_result_urls) && record.matching_result_urls.present?
   }
 end
--- a/spec/presenters/keyword_presenter_spec.rb
+++ b/spec/presenters/keyword_presenter_spec.rb
@@ -36,5 +36,16 @@ RSpec.describe KeywordPresenter do
         expect(result).to be_nil
       end
     end
+
+    context 'given result urls contain a word vpn' do
+      it 'returns matching_result_urls with 2 urls https://www.topvpn.com and https://www.nordvpn.com' do
+        result_urls = ['https://www.topvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+        keyword = Fabricate(:keyword, result_urls: result_urls)
+        filter_params = { word: 'vpn', match_at_least: '1' }
+        result = described_class.new(keyword, filter_params).matching_result_urls
+
+        expect(result).to eq(['https://www.topvpn.com', 'https://www.nordvpn.com'])
+      end
+    end
   end
 end
--- a/spec/presenters/keyword_presenter_spec.rb
+++ b/spec/presenters/keyword_presenter_spec.rb
@@ -11,7 +11,7 @@ RSpec.describe KeywordPresenter do
         filter_params = { adwords_url_contains: 'vpn' }
         result = described_class.new(keyword, filter_params).matching_adword_urls
 
-        expect(result).to eq(['https://www.thetopvpn.com', 'https://www.nordvpn.com'])
+        expect(result).to contain_exactly('https://www.thetopvpn.com', 'https://www.nordvpn.com')
       end
     end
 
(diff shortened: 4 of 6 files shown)


spec/queries/keywords_query_spec.rb
```diff
@@ -38,6 +38,30 @@
...
+          result_urls = ['https://www.thetopvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+          keyword = Fabricate(:keyword, result_urls: result_urls, user: user)
+          filter_params = { word: 'vpn', match_at_least: '1' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to eq [keyword]
+        end
+      end
+
+      context 'given the user search for vpn word e and match at least 2' do
```
blyscuit: Now that we gives priority to `adwords_url_contains` over `match_at_least`, I think we should add spec with context 'given the user search for adwords, word and match at least`.

Also, on that note, we should test with `word` without `match at least` and `match at a least` without `word` too 🙏 .

blyscuit (commented):
One comment, looks good.

spec/queries/keywords_query_spec.rb
phongvhd93: solved in 33a43db

blyscuit: How about this one to test priority declared in `keyword_query` 😄 
```
          filter_params = { adwords_url_contains: 'word' word: 'vpn', match_at_least: '2' }
# should return adwords result
```

app/queries/keywords_query.rb
```diff
@@ -9,14 +9,28 @@ def initialize(scope, filter_params)
   end
 
   def call
-    return scope unless defined?(filter_params[:adwords_url_contains]) && filter_params[:adwords_url_contains]
-
-    filter_ads_top_urls(filter_params[:adwords_url_contains])
+    if filter_params[:adwords_url_contains]
```
byhbt: Just minor suggestion, we can return if early to reduce the nested level.

```suggestion    
      return filter_ads_top_urls(filter_params[:adwords_url_contains]) if filter_params[:adwords_url_contains]
      return filter_result_urls(filter_params[:word], filter_params[:match_at_least]) if filter_params[:word] && filter_params[:match_at_least]

      scope
    end
```

spec/presenters/keyword_presenter_spec.rb
```diff
@@ -36,5 +36,16 @@
...
     end
+
+    context 'given result urls contain a word vpn' do
+      it 'returns matching_result_urls with 2 urls https://www.topvpn.com and https://www.nordvpn.com' do
+        result_urls = ['https://www.topvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+        keyword = Fabricate(:keyword, result_urls: result_urls)
+        filter_params = { word: 'vpn', match_at_least: '1' }
+        result = described_class.new(keyword, filter_params).matching_result_urls
+
+        expect(result).to eq(['https://www.topvpn.com', 'https://www.nordvpn.com'])
```
byhbt:
```suggestion
        expect(result).to contain_exactly('https://www.topvpn.com', 'https://www.nordvpn.com')
```

spec/queries/keywords_query_spec.rb
```diff
@@ -30,12 +30,66 @@
...
+    context 'given the user search for result_urls' do
+      context 'given the user search for a word vpn and match at least 1' do
+        it 'return a keyword' do
+          user = Fabricate(:user)
+          result_urls = ['https://www.thetopvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+          keyword = Fabricate(:keyword, result_urls: result_urls, user: user)
+          filter_params = { word: 'vpn', match_at_least: '1' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to eq [keyword]
```
byhbt:
```suggestion
          expect(keywords).to contain_exactly(word)
```

spec/queries/keywords_query_spec.rb
```diff
@@ -30,12 +30,66 @@
...
 
-          expect(keywords.count).to eq 2
+          expect(keywords).to eq Keyword
+        end
+      end
+    end
+
+    context 'given the user search for result_urls' do
+      context 'given the user search for a word vpn and match at least 1' do
+        it 'return a keyword' do
```
byhbt:
```suggestion
        it 'returns a keyword' do
```

spec/requests/api/v1/keywords_controller_spec.rb
```diff
@@ -159,6 +159,160 @@
...
       end
+
+      context 'given the user search for a specific word in result_urls' do
+        context 'given the user search for word vpn with match_at_least 1' do
+          it 'returns keywords contains the matched urls' do
+            user = Fabricate(:user)
+            result_urls = ['https://www.thetopvpn.com', 'https://www.nordvpn.com', 'https://www.vnexpress.net']
+            Fabricate(:keyword, result_urls: result_urls, user: user)
+            params = { filter: { word: 'vpn', match_at_least: '1' }, page: 1, per_page: 1 }
+            get api_v1_keywords_path, params: params, headers: create_token_header(user)
```
byhbt, line 171: Add one empty line before the request, same for other test below.

```suggestion

            get api_v1_keywords_path, params: params, headers: create_token_header(user)
```

app/presenters/keyword_presenter.rb
```diff
@@ -12,6 +12,18 @@ def matching_adword_urls
     return unless defined?(filter_params[:adwords_url_contains]) && filter_params[:adwords_url_contains]
```
byhbt: I am not sure why do we need to check `defined?` here? If you would like to check `nil` you can use `present?` instead. 

https://ruby-doc.org/docs/keywords/1.9/Object.html#method-i-defined-3F

phongvhd93: I got NoMethodError when trying to I used `present?` or `any?` when the `filter_params` is nil. So trying to access the  param `adwords_url_contains` in a nil object will throw an error.

malparty: You can avoid this with a default value: `def initialize(keyword, filter_params = {})`.

```suggestion
    return unless filter_params[:adwords_url_contains]
```

app/presenters/keyword_presenter.rb
```diff
@@ -12,6 +12,18 @@ def matching_adword_urls
...
 
     keyword.ads_top_urls.select { |item| item.downcase.include?(filter_params[:adwords_url_contains].downcase) }
+  rescue NoMethodError
+    nil
+  end
+
+  def matching_result_urls
+    return unless filter_params[:word] && filter_params[:match_at_least]
```
malparty:
```suggestion
    return if filter_params[:word].blank? || filter_params[:match_at_least].blank?
```

Composed `unless` can quickly become: 🤯 . So when more than one precicate, let's use a normal `if` instead :)

app/presenters/keyword_presenter.rb
```diff
@@ -12,6 +12,18 @@ def matching_adword_urls
...
+
+    keyword.result_urls.select { |item|
```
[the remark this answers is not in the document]
malparty: use `do ... end` instead 🙈

app/presenters/keyword_presenter.rb
```diff
@@ -12,6 +12,18 @@ def matching_adword_urls
     return unless defined?(filter_params[:adwords_url_contains]) && filter_params[:adwords_url_contains]
 
     keyword.ads_top_urls.select { |item| item.downcase.include?(filter_params[:adwords_url_contains].downcase) }
```
malparty:
```suggestion
    adwords_url_contains = filter_params[:adwords_url_contains].downcase
    keyword.ads_top_urls.select { |item| item.downcase.include?(adwords_url_contains) }
```

Sorry I missed that one in the previous PRs 🙈

malparty: (for performance reason, let's not downcase the same string multiple times in a loop)

app/presenters/keyword_presenter.rb
```diff
@@ -12,6 +12,18 @@ def matching_adword_urls
...
+  rescue NoMethodError
+    nil
+  end
+
+  def matching_result_urls
+    return unless filter_params[:word] && filter_params[:match_at_least]
+
+    keyword.result_urls.select { |item|
+      item.downcase.match("\\w*(#{filter_params[:word]}.*){#{filter_params[:match_at_least]},}")
```
malparty: Are the filters already down-cased?
if not, we must do that outside from the select loop 😇

app/queries/keywords_query.rb
```diff
@@ -9,14 +9,28 @@ def initialize(scope, filter_params)
...
+    scope
   end
 
   private
 
   def filter_ads_top_urls(word)
     scope.where("array_to_string(ads_top_urls, '||') ILIKE ?", "%#{word}%")
   end
+
+  def filter_result_urls(word, number_of_matches)
```
```suggestion
  def filter_ads_top_urls
    word = filter_params[:adwords_url_contains]&.downcase
    
    return unless word
    
    scope.where("array_to_string(ads_top_urls, '||') ILIKE ?", "%#{word}%")
  end

  def filter_result_urls
    # same here, include the condition logic and the filter param processing
```

Let's leverage OOP to make the conditions and calls above less complex.

Then the call method becomes much easier to read: it just calls each filtering method in the expected order. The logic for each filtering is then contained inside 1 method (and not spread between `call` and the filtering method).

app/serializers/keyword_serializer.rb
```diff
@@ -4,9 +4,14 @@ class KeywordSerializer < ApplicationSerializer
...
-             :ads_top_urls
+             :ads_top_urls,
+             :result_urls
 
   attribute :matching_adword_urls, if: proc { |record|
     defined?(record.matching_adword_urls) && record.matching_adword_urls.present?
   }
+
+  attribute :matching_result_urls, if: proc { |record|
+    defined?(record.matching_result_urls) && record.matching_result_urls.present?
```
malparty:
```suggestion
    record&.matching_result_urls.present?
```
does this work?

spec/queries/keywords_query_spec.rb
```diff
@@ -4,7 +4,7 @@
 
 RSpec.describe KeywordsQuery, type: :query do
   describe '#call' do
-    context 'given user has keyword' do
+    context 'given the user search for ads_top_urls' do
```
malparty:
```suggestion
    context 'given the user searches for ads_top_urls' do
```

spec/queries/keywords_query_spec.rb
```diff
@@ -30,12 +30,66 @@
...
           keywords = described_class.new(Keyword, nil).call
 
-          expect(keywords.count).to eq 2
+          expect(keywords).to eq Keyword
+        end
+      end
+    end
+
+    context 'given the user search for result_urls' do
+      context 'given the user search for a word vpn and match at least 1' do
```
malparty:
```suggestion
      context 'given the user searches for a word vpn and match at least 1' do
```

malparty: That said, a "query" object, does not care about the user... 🤔 
Should we use more "argument-driven" descriptions? Like:

```suggestion
      context 'given filters with the word `vpn` and match at least `1`' do
```

malparty: Now this filter is "allowed" to be used by a bot, by a rake task or a background job if needed 😇

spec/queries/keywords_query_spec.rb
```diff
@@ -30,12 +30,66 @@
...
+          expect(keywords).to eq Keyword
+        end
+      end
+
+      context 'given user provides match_at_least parameter and leaves word parameter nil' do
+        it 'returns Keyword scope' do
+          filter_params = { match_at_least: 'vpn' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to eq Keyword
```
malparty: While this works, it's quite implicit. Should we make the `.all` scope explicit?

```suggestion
        it 'returns all the keywords' do
          filter_params = { match_at_least: 'vpn' }
          keywords = described_class.new(Keyword, filter_params).call

          expect(keywords).to eq Keyword.all
```

malparty (changes requested):

blyscuit (approved):

spec/queries/keywords_query_spec.rb
```diff
@@ -4,12 +4,11 @@
 
 RSpec.describe KeywordsQuery, type: :query do
   describe '#call' do
-    context 'given user has keyword' do
+    context 'given the user searchs for ads_top_urls' do
```
malparty:
```suggestion
    context 'given the user searches for ads_top_urls' do
```

did I miss that `e` in the previous suggestion 🙈

spec/queries/keywords_query_spec.rb
```diff
@@ -30,14 +28,78 @@
...
+          expect(keywords).to be_empty
+        end
+      end
+
+      context 'given word parameter is vpn and match_at_least parameter nil' do
+        it 'returns Keyword scope' do
+          filter_params = { word: 'vpn' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to eq Keyword
```
malparty:
```suggestion
        it 'returns all the keywords' do
          filter_params = { word: 'vpn' }
          keywords = described_class.new(Keyword, filter_params).call

          expect(keywords).to eq Keyword.all
```

spec/queries/keywords_query_spec.rb
```diff
@@ -30,14 +28,78 @@
...
+          keyword = Fabricate(:keyword, result_urls: result_urls)
+          filter_params = { word: 'vpn', match_at_least: '2' }
+          keywords = described_class.new(Keyword, filter_params).call
+
+          expect(keywords).to contain_exactly(keyword)
+        end
+      end
+
+      context 'given there is no matched result' do
+        it 'returns empty' do
```
malparty:
```suggestion
        it 'is empty' do
```

or it returns an empty scope

spec/queries/keywords_query_spec.rb
```diff
@@ -30,14 +28,78 @@
       end
 
       context 'given filter_params is nil' do
-        it 'returns all keywords' do
-          user = Fabricate(:user)
-          Fabricate.times(2, :keyword, user: user)
+        it 'returns Keyword scope' do
```
malparty:
```suggestion
        it 'returns all keywords' do
```

spec/queries/keywords_query_spec.rb
```diff
@@ -30,14 +28,78 @@
...
           keywords = described_class.new(Keyword, nil).call
 
-          expect(keywords.count).to eq 2
+          expect(keywords).to eq Keyword
```
malparty:
```suggestion
          expect(keywords).to eq Keyword.all
```

malparty (approved):
Some last minor suggestions for the tests (about consistency).

As an insight (but not needed to change) for the Query object, you can leverage OOP by moving the `conditions` of each filter inside the filtering method.
This way the `call` method only focuses on calling each filters in the right order. Then each filter method contain the whole logic about it's cases:

```

  def call
    filter_ads_top_urls
    filter_result_urls

    scope
  end

private

  def filter_ads_top_urls
    return unless filter_params[:adwords_url_contains]

    word = filter_params[:adwords_url_contains]&.downcase

    # ...... (assign `scope = ...`)
  end

  def filter_result_urls
    return if filter_params[:word].blank? || filter_params[:match_at_least].blank?

    word = filter_params[:word]

    # ...... (assign `scope = ...`)
  end
```

(the `rescue NoMethodError` could also be handled in the filter method in need only).

byhbt (approved):

